Issue #34, opened 2024-09-03
Brotli compression


SachaMorard:
**Is your feature request related to a problem? Please describe.**

When a request arrives at the Edgee proxy, it is forwarded to an upstream server. Edgee forwards the client's http headers, and if the upstream server supports compression, a gzip/deflate/brotli response can be sent. 
Edgee already supports Gzip and Deflate compression, but not Brotli.

This means that if a Brotli response is returned by an upstream server, Edgee will not be able to launch its computational functions (data collection, ab tests...).

**Describe the solution you'd like**
Add a brotli library (the most performant possible) and implement within the entrypoint.rs logic

PR #39, opened 2024-09-04
Brotli encoding

CLEMENTINATOR:
### Checklist

### Description of Changes

handle brotli ("br") encoding.

Takes around 120ms on analytics-with-edgee.html page (including network time), but only 16ms if proxy is built in release.

### Related Issues

Fixes issue #34

### DIFF
--- a/src/entrypoint.rs
+++ b/src/entrypoint.rs
@@ -24,6 +24,7 @@ use tracing::{debug, error, warn};
 use crate::{tools::{edgee_cookie, path}, data_collection::{self, Session}, destinations, config};
 use crate::tools::real_ip::Realip;
 use crate::{data_collection::Payload, html};
+use brotli::{Decompressor, CompressorWriter};
 
 mod incoming_context;
 mod proxy_context;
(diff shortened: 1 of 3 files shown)


src/entrypoint.rs
```diff
@@ -452,8 +463,19 @@ async fn handle_request(request: http::Request<Incoming>, remote_addr: SocketAdd
...
-                // todo brotli
-                Some(_) | None => new_body.into(),
+                Some("br") => { // handle brotli encoding
+                    // q: quality (range: 0-11), lgwin: window size (range: 10-24)
+                    let mut encoder  = CompressorWriter::new(Vec::new(), 4096, 11, 24);
+                    encoder.write_all(new_body.as_bytes())?;
+                    encoder.flush()?;
+                    encoder.into_inner()
+                }
+                Some(enc) => { // encoding not supported
```
SachaMorard, line 473: I didn't test it, but I think you can skip this step, because you already returned the response if the encoding is not supported (line 348)

SachaMorard (changes requested):

src/entrypoint.rs
CLEMENTINATOR: yeah we should not get here, fixing

SachaMorard (approved):

PR #84, opened 2024-10-01
feat: async http calls from component and bugfixes

SachaMorard:
### Checklist

### Description of Changes

This pull request introduces the implementation of data collection components using WebAssembly (Wasm) and enhances the payload processing logic. The changes are primarily focused on the `src/proxy/compute/data_collection/components.rs` file.

- Implemented the initialization of Wasm components for data collection.
- Added the send_data_collection function to handle data collection http calls asynchronously.
- Introduced the anonymize_ip function to anonymize client IP addresses.
- Defined the HostView struct to manage Wasm context and resources.

### Related Issues

no issue

### DIFF
--- a/src/proxy/compute/data_collection/components.rs
+++ b/src/proxy/compute/data_collection/components.rs
@@ -4,12 +4,13 @@ static WASM_LINKER: OnceCell<wasmtime::component::Linker<HostView>> = OnceCell::
 static WASM_ENGINE: OnceCell<wasmtime::Engine> = OnceCell::const_new();
 static WASM_COMPONENTS: OnceCell<HashMap<&str, Component>> = OnceCell::const_new();
 
-use std::{collections::HashMap, str::FromStr};
-
 use crate::config::config;
 use crate::proxy::compute::data_collection::payload::{EventType, Payload};
 use exports::provider;
 use http::{HeaderMap, HeaderName, HeaderValue};
+use std::net::IpAddr;
+use std::time::Duration;
+use std::{collections::HashMap, str::FromStr};
 use tokio::sync::OnceCell;
 use tracing::{error, info};
 use wasmtime::component::Component;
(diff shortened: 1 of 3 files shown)


src/proxy/compute/data_collection/components.rs
```diff
@@ -45,8 +46,240 @@ pub async fn send_data_collection(p: &Payload) -> anyhow::Result<()> {
     let engine = WASM_ENGINE.get().unwrap();
     let linker = WASM_LINKER.get().unwrap();
     let mut store = wasmtime::Store::new(engine, HostView::new());
+
+    // clone the payload to be able to move it to the async thread
+    let p = p.clone();
+    let payload = provider::Payload {
```
KokaKiwi, line 52: I feel like this while block could gain from being split and implemented into `payload.rs` as a `impl From<provider::Payload> for Payload`, same for `PageEvent` and others from the same file, so we could split this whole block into smaller and simpler blocks.
Not specifically in this PR but in a future "cleanup" PR maybe?

KokaKiwi (approved):
I feel like the whole "data collection" code should be the one being put in an async task instead of all the small bits (HTTP requests, etc.) since it's literally a "side-effect"
But that would probably require a lot more refactoring (cf. comment about cleaning up the payload conversion code) which maybe should be done in a dedicated cleanup PR?

KokaKiwi:
Oh also the formatting seems odd, as in: there's a lot of code reformatting which has been done despite the `cargo fmt` previously done in #80, is that something from RustRover doing its own formatting rules?

src/proxy/compute/data_collection/components.rs
SachaMorard: Yes, we'll do it in a future PR

SachaMorard:
> Oh also the formatting seems odd, as in: there's a lot of code reformatting being done in this PR despite the `cargo fmt` previously done in #80, is that something from RustRover doing its own formatting rules?

I don't think so. I use cargo fmt manually to be sure the last pushed commit is correctly formatted

KokaKiwi:
> I don't think so. I use cargo fmt manually to be sure the last pushed commit is correctly formatted

Uh. Do you maybe have a custom `rustfmt.toml` config file? Or maybe a different Rust version?

Issue #40, opened 2024-09-04
Brotli compression performance improvement


SachaMorard:
**Is your feature request related to a problem? Please describe.**
Brotli compression is slower than other encoding algorithms 

Edgee adds a header to the response which indicates the compute duration (`x-edgee-compute-duration`)
When an upstream server responds a gzip compressed response, the compute duration on my local env is okay (~10ms)
When an upstream server responds a brotli compressed response, the compute duration on my local env is too slow (~120ms)

The slowdown seems to originate from this part:
```rust
Some("br") => { // handle brotli encoding
    // q: quality (range: 0-11), lgwin: window size (range: 10-24)
    let mut encoder  = CompressorWriter::new(Vec::new(), 4096, 11, 24);
    encoder.write_all(new_body.as_bytes())?;
    encoder.flush()?;
    encoder.into_inner()
}
```

**Describe the solution you'd like**
- Test if brotli compression settings can help
- Look at the strategies and solutions used by other proxies 
- Test other libraries

asolimando:
Here are some figures for `x-edgee-compute-duration` on my machine with different setups: 
- no compression: ~3ms
- brotli (quality 5): ~9 ms
- brotli (quality 11): ~101 ms <-- that's the current hard-coded value
- gzip: ~9 ms

So already by lowering the quality to `5` we have a performance closer to that of `gzip`, of course a couple of runs without proper benchmarking isn't conclusive, but it already suggests that there is room for improvement by changing some of the default parameters (with #97 in place it would probably be easier to have better figures).

Regarding what other reverse proxies do:
Nginx allows to tune several brotli parameters, including [quality/compression level](https://github.com/google/ngx_brotli?tab=readme-ov-file#brotli_comp_level), [buffer size](https://github.com/google/ngx_brotli?tab=readme-ov-file#brotli_buffers), [window size](https://github.com/google/ngx_brotli?tab=readme-ov-file#brotli_window) and the [minimum payload size](https://github.com/google/ngx_brotli?tab=readme-ov-file#brotli_min_length).

Additionally, not all MIME types equally benefit from compression, so we have an additional parameter for the [MIME types](https://github.com/google/ngx_brotli?tab=readme-ov-file#brotli_types) for which one wants brotli encoding to be activated.

In summary, a reasonable starting point is to lower the compression quality parameter, and possibly add all or some of these configuration knobs to the Edgee configuration to make it tunable.

PR #103, opened 2024-10-17
perf(proxy): Improve compression performance

KokaKiwi:
### Checklist

### Description of Changes

- Add base for tower-based request handling
- Defer response compression to `tower-http` layers
- Add `compress` config field for backends to enable/disable compression between edgee and the endpoint (disabled by default)

### References

- Benches: https://gist.github.com/KokaKiwi/5b2f59fc0cf322950f149652361cc993

### Possibles improvements

- Maybe we should enable by default compression between edgee and backend endpoint?
	- Should we enable compression even when client did not request compressed response?

### Related Issues

- #40


src/config/config.rs
```diff
@@ -185,8 +184,6 @@ pub struct ComputeConfiguration {
     pub behind_proxy_cache: bool,
     #[serde(default = "default_max_decompressed_body_size")]
     pub max_decompressed_body_size: usize,
-    #[serde(default = "default_max_compressed_body_size")]
-    pub max_compressed_body_size: usize,
```
SachaMorard, line 189: @KokaKiwi why removing this configuration line?
It is important for the computation process and is not linked to the response compression.

src/proxy/context/proxy.rs
```diff
@@ -57,6 +57,9 @@ impl<'a> ProxyContext<'a> {
             );
         }
 
+        // We want uncompressed response from endpoint since we already do the compression
+        incoming_headers.remove("accept-encoding");
```
SachaMorard, line 61: I would prefer to make this optional. 
Because in some cases, it could have a huge impact on the bandwidth cost of the user

src/proxy/controller/controller.rs
```diff
@@ -173,6 +170,8 @@ pub fn build_response(mut parts: http::response::Parts, body: Bytes) -> Response
     // Update Content-Length header to correct size
     parts.headers.insert("content-length", body.len().into());
 
+    parts.headers.remove("transfer-encoding");
```
SachaMorard, line 173: Why remove this response header?

SachaMorard (changes requested):
Your PR seems very interesting, but I'd keep the decompression process as it is, or at least make it optional. In some cases, the user might want to keep compressed exchanges between the backend and Edgee to reduce bandwidth costs...

This PR is quite important, and in this case, I recommend not “polluting” it with refacto lines, as this makes it easier to read.

src/proxy/controller/controller.rs
KokaKiwi: It was a header i removed during dev because i'm using `miniserve` to serve the demo-html with compression and it was adding this header with value `chunked` which conflicted with other `content-length` (as per [the specs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Transfer-Encoding#chunked)) and `hyper` would then throw an error.
I wanted to investigate more this "corner case" before actually committing a fix (if needed) and i forgot to remove this temporary line when creating the PR

src/config/config.rs
```diff
@@ -158,6 +158,8 @@ pub struct BackendConfiguration {
     pub default: bool,
     pub address: String,
     pub enable_ssl: bool,
+    #[serde(default)]
+    pub compress: bool,
```
SachaMorard, line 162: 👍

src/proxy/proxy.rs
```diff
@@ -143,6 +142,43 @@ pub async fn handle_request(
...
+                        body.collect()
+                            .await
+                            .map_err(|err| anyhow::anyhow!(err))?
+                            .to_bytes()
+                    }
+                }
+            };
+
             // Only proxy in some cases
             match do_only_proxy(request.get_method(), &response_body, &response) {
```
SachaMorard, line 183: `do_only_proxy` is now happening after the decompression... 
I'm not sure about its side effects. 
For example, if the backend serves a CSS resource, we don't need to decompress the response at all; we can simply serve it as is. With this implementation, we decompress each backend response, right?

SachaMorard (changes requested):

SachaMorard:
@KokaKiwi it's really good! 👏👏
I just have a concern about the decompression timing. Let's have a talk about it

KokaKiwi:
> @KokaKiwi it's really good! 👏👏 I just have a concern about the decompression timing. Let's have a talk about it

It was not a trivial fix but now it should decompress the body only when needed, also the fix made the code a bit cleaner with the addition of `ProxyBody` :)

SachaMorard (approved):

PR #116, opened 2024-10-31
feat: component debug feature

SachaMorard:
### Checklist

### Description of Changes

Feature to debug a component easily
Developing a component is not easy, especially if we can't see anything. 

Now, we can launch edgee with a specific option `--debug-component=<component name>` and edgee will only print the content of the requests sent by the component, and the result of the analytics provider.

### Related Issues

No issue

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -18,14 +18,29 @@ struct Options {
 
     #[arg(short = 'f', long = "config", env = "EDGEE_CONFIG_PATH")]
     config_path: Option<PathBuf>,
+
+    #[arg(
+        short = 'c',
+        long = "debug-component",
+        help = "Launch Edgee and log only the specified component requests and responses to debug.",
+        id = "COMPONENT_NAME"
+    )]
+    debug_component: Option<String>,
 }
 
 #[tokio::main]
 async fn main() {
     let options = Options::parse();
 
-    config::config::init(options.config_path.as_deref());
-    logger::init(options.log_format);
+    config::config::init(&options);
+    // if debug_component is set, we only want to log the specified component. We change the options.log_format to do it.
+    let mut log_filter = None;
+    if options.debug_component.is_some() {
+        // We disable all logs because component will print things to stdout directly
+        log_filter = Some("[none]".to_string());
+    }
+
+    logger::init(options.log_format, log_filter);
     proxy::compute::data_collection::components::init();
 
     tokio::select! {
--- a/src/main.rs
+++ b/src/main.rs
@@ -37,7 +37,7 @@ async fn main() {
     let mut log_filter = None;
     if options.debug_component.is_some() {
         // We disable all logs because component will print things to stdout directly
-        log_filter = Some("[none]".to_string());
+        log_filter = Some("off".to_string());
     }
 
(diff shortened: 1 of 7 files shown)


alexcasalboni:
Looks very useful 👌  I'd recommend adding a short section (or even just a paragraph) on the README as well

src/main.rs
```diff
@@ -18,14 +18,29 @@ struct Options {
...
     let options = Options::parse();
 
-    config::config::init(options.config_path.as_deref());
-    logger::init(options.log_format);
+    config::config::init(&options);
+    // if debug_component is set, we only want to log the specified component. We change the options.log_format to do it.
+    let mut log_filter = None;
+    if options.debug_component.is_some() {
+        // We disable all logs because component will print things to stdout directly
+        log_filter = Some("[none]".to_string());
```
KokaKiwi, line 40: If we want to disable all logs, `off` should be used, as defined by [`tracing_subscriber::filter::LevelFilter::OFF`](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/struct.LevelFilter.html#associatedconstant.OFF) value.

Also, since this value seems to be (for now maybe?) only a two-state value (on/off), shouldn't the type be a simple `bool`?

```suggestion
        log_filter = Some("off".to_string());
```

src/proxy/compute/data_collection/data_collection.rs
```diff
@@ -104,14 +104,11 @@ pub async fn process_from_html(
...
-    tokio::spawn(
-        async move {
-            if let Err(err) = components::send_data_collection(&events).await {
-                warn!(?err, "failed to send data collection payload");
-            }
+    tokio::spawn(async move {
+        if let Err(err) = components::send_data_collection(&events).await {
+            warn!(?err, "failed to send data collection payload");
         }
-        .in_current_span(),
```
KokaKiwi, line 113: The spawned task should keep the attached span since it links the other spans created within to the parent span, thus helping grouping events in the same "trace"

src/proxy/compute/data_collection/data_collection.rs
```diff
@@ -198,14 +195,11 @@ pub async fn process_from_json(
...
```
KokaKiwi, line 207: Same as above

src/proxy/compute/data_collection/data_collection.rs
SachaMorard: I created a dedicated span for the component part, which is simpler when you want to debug only one component. That's why I decoupled it like that.

KokaKiwi: Spans are arranged in a tree in order to define a "trace" (with a root span which is considered as the "trace root"), it doesn't conflict with creating other spans within which can also be used to filter logs (for example with `[component{name="your-component-name"}]` filter)
It's better to link spans as much as possible in order to create meaningful traces when doing observability (especially using OpenTelemetry)

SachaMorard: Oh wow, I didn't know that. Tracing is really powerful. Thank you @KokaKiwi, I'll put `.in_current_span()` back in the next PR

Issue #118, opened 2024-11-01
Get all the User-Agent Client Hints


SachaMorard:
[**Client hints**](https://developer.mozilla.org/en-US/docs/Web/HTTP/Client_hints) are a set of [HTTP request header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers) fields that a server can proactively request from a client to get information about the device, network, user, and user-agent-specific preferences. The server can determine which resources to send, based on the information that the client chooses to provide.<br>Client hints are broadly divided into high and low entropy hints.

* Low Entropy
* High Entropy

## Low Entropy 

The low entropy hints are those that don't give away much information that might be used to create a fingerprinting for a user. They may be sent by default on every client request, irrespective of the server `Accept-CH` response header, depending on the permission policy. These hints include: [`Sec-CH-UA`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Sec-CH-UA), [`Sec-CH-UA-Mobile`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Sec-CH-UA-Mobile), [`Sec-CH-UA-Platform`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Sec-CH-UA-Platform).

Edgee is already gather those low entropy client hints in the request headers.

## High Entropy

High entropy client hints are rarely sent in headers.

We often miss the following headers:

* [Sec-Ch-Ua-Arch](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Sec-CH-UA-Arch) (user_agent_architecture)
* [Sec-Ch-Ua-Bitness](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Sec-CH-UA-Bitness) (user_agent_bitness)
* [Sec-Ch-Ua-Full-Version-List](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Sec-CH-UA-Full-Version-List) (user_agent_full_version_list)
* [Sec-Ch-Ua-Model](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Sec-CH-UA-Model) (user_agent_model)
* [Sec-Ch-Ua-Platform-Version](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Sec-CH-UA-Platform-Version) (os_version)

[...]

PR #119, opened 2024-11-02
feat: get all client hints

SachaMorard:
### Checklist

### Description of Changes

Get all client hints
- new js SDK that grap the high entropy client hints
- new convert_dict function that better converts any properties into string
- better debugging from component calls
- bump wasmtime 
- get all high entropy client hints and put them into the cookie and the payload

### Related Issues

#118


src/proxy/compute/data_collection/components/convert.rs
```diff
@@ -134,8 +135,21 @@ impl From<payload::Session> for provider::Session {
...
+            continue;
+        }
+        let s = v.as_str();
+        if let Some(s) = s {
+            vec.push((k.clone(), s.to_string()));
+        } else {
+            vec.push((k.clone(), v.to_string()));
+        }
+    }
+    vec
```
KokaKiwi, line 154: You can use the `let Some($pat) = $value else { $code };` pattern to avoid unwrapping around (both generate kinda the same assembly at the end but the "let ... else ..." pattern is a really nice syntax to benefit the full type-safety of Rust :slightly_smiling_face:)

Also i would suggest using the iterator-based way of converting stuff using filter and map, but that's mainly a "procerural vs functional" styling issue :smile: 

```suggestion
    use serde_json::Value;

    let Some(dict) = dict else {
        return Vec::new();
    };

    dict.into_iter()
        .filter(|(_, value)| !(value.is_array() || value.is_object()))
        .map(|(k, v)| {
            let value = if let Value::String(ref v) = v {
                v.clone()
            } else {
                v.to_string()
            };

            (k, value)
        })
        .collect()
```

PR #144, opened 2024-12-11
feat: products properties

SachaMorard:
### Checklist

### Description of Changes

For now, Edgee data collection components can only collect standard events (pageviews, identify, simple custom event). To be complete, a user has to be able to send e-commerce data to his analytics providers. 
Following the [GA4 e-commerce spec](https://developers.google.com/analytics/devguides/collection/ga4/ecommerce?hl=en&client_type=gtag) and [Segment doc](https://segment.com/docs/connections/spec/ecommerce/v2/), I added the interpretation of a `products` array in the `data.properties` of a track event. 

### Related Issues

No issue

### DIFF
--- a/crates/server/src/proxy/compute/data_collection/components/convert.rs
+++ b/crates/server/src/proxy/compute/data_collection/components/convert.rs
@@ -20,6 +20,7 @@ impl From<payload::Event> for provider::Event {
             event_type: value.event_type.into(),
             data,
             context: value.context.unwrap_or_default().into(),
+            consent: provider::Consent::Pending,
         }
     }
 }
--- a/crates/server/src/proxy/compute/data_collection/components/convert.rs
+++ b/crates/server/src/proxy/compute/data_collection/components/convert.rs
@@ -46,7 +46,6 @@ impl From<payload::Page> for provider::PageData {
             search: value.search.unwrap_or_default(),
             referrer: value.referrer.unwrap_or_default(),
             properties: convert_properties(value.properties.clone()),
-            products: convert_products(value.properties.clone()),
         }
     }
 }
--- a/crates/server/src/proxy/compute/data_collection/components/convert.rs
+++ b/crates/server/src/proxy/compute/data_collection/components/convert.rs
@@ -1,4 +1,5 @@
 use std::collections::HashMap;
+use tracing::warn;
 
 use super::exports::provider;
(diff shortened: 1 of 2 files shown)


crates/server/src/proxy/compute/data_collection/components/convert.rs
```diff
@@ -155,3 +159,50 @@ fn convert_dict(dict: Option<HashMap<String, serde_json::Value>>) -> Vec<(String
...
+    let Some(dict) = properties else {
+        return Vec::new();
+    };
+    
+
+    // if the key is products, then we need to convert the value to a list of tuples
+    if let Some(products) = dict.get("products") {
+        // if products is not an array, return an empty vector
+        if !products.is_array() {
+            return Vec::new();
```
alexcasalboni, line 177: should we log this somehow so that we can more easily debug this scenario? if `products` is meant to always be an array, we should make sure the developer has a way to notice that mistake instead of silently failing

crates/server/src/proxy/compute/data_collection/components/convert.rs
```diff
@@ -155,3 +159,50 @@ fn convert_dict(dict: Option<HashMap<String, serde_json::Value>>) -> Vec<(String
...
+        }
+
+        let mut results: Vec<Vec<(String, String)>> = Vec::new();
+        let items = products.as_array().unwrap();
+        items.iter().for_each(|product| {
+            // if product is not an object, go to the next product
+            if !product.is_object() {
+                return;
```
alexcasalboni, line 185: same here - if the product object isn't valid the developer should have a way to notice the mistake and correct their code (or at least we should get notified and let them know)

crates/server/wit/protocols.wit
```diff
@@ -32,6 +32,7 @@ world data-collection {
             search: string,
             referrer: string,
             properties: dict,
+            products: list<dict>,
```
alexcasalboni, line 35: I don't know if it helps at this stage, but we could start using gates/annotations such as `@unstable` and `@since(version = 0.2.3)` (the official recommendation is to start with `@unstable` and then convert it to `@since` once the new feature is solid)

SachaMorard: +1 Let's do this when the component registry is live!

alexcasalboni (approved):

PR #148, opened 2024-12-13
fix: referer for data collection call

SachaMorard:
### Checklist

### Description of Changes

The referer header was wrongly defined on the data collection calls. 
It has to be inserted like an analytics client-side SDK could do

### Related Issues

List related issues here


KokaKiwi (approved):

alexcasalboni:
Should this be updated here too? https://github.com/edgee-cloud/sdk/blob/main/js/analytics/context.js#L137

Or that's ok?

SachaMorard:
> Should this be updated here too? https://github.com/edgee-cloud/sdk/blob/main/js/analytics/context.js#L137
> 
> 
> 
> Or that's ok?

No need. It was only a http header problem on the data collection calls.

PR #149, opened 2024-12-13
cli: Add auth command

KokaKiwi:
### Checklist

### Description of Changes

- Add a new crate `edgee-components` to house the code related to components authoring, building and publishing (and associated utilities)
- Add the new CLI command `auth` with two subcommands:
  * `login`: Ask for an API token which can be created on the console and save it into a credentials file after validation
  * `whoami`: Ask the API for who is the currently logged user, if any

### Usage example

```shell
$ edgee auth login
> Enter Edgee API token (you can create one at https://www.edgee.cloud/<username>/settings/tokens): ******
Logged as Julianne Hervier (julianne@edgee.cloud)
```

### Related Issues

- EDGEE-230

### DIFF
--- a/crates/cli/src/commands/auth/login.rs
+++ b/crates/cli/src/commands/auth/login.rs
@@ -15,9 +15,7 @@ pub async fn run(_opts: Options) {
     }
 
     let api_token = Password::new("Enter Edgee API token (press Ctrl+R to toggle input display):")
-        .with_help_message(
-            "You can create one at https://www.edgee.cloud/<username>/settings/tokens",
-        )
+        .with_help_message("You can create one at https://www.edgee.cloud/me/settings/tokens")
         .with_display_mode(PasswordDisplayMode::Masked)
         .with_display_toggle_enabled()
         .without_confirmation()
--- a/crates/cli/src/commands/auth/login.rs
+++ b/crates/cli/src/commands/auth/login.rs
@@ -1,0 +1,30 @@
+setup_command! {}
+
+pub async fn run(_opts: Options) {
+    use inquire::{Confirm, Text};
+
+    use edgee_components::auth::Credentials;
+
+    let mut creds = Credentials::load().unwrap();
+
+    let confirm_overwrite =
+        Confirm::new("An API token is already present, do you want to overwrite it?")
+            .with_default(false);
+    if creds.api_token.is_some() && !confirm_overwrite.prompt().unwrap() {
+        return;
+    }
+
+    let api_token = Text::new("Enter Edgee API token:").prompt().unwrap();
+    creds.api_token.replace(api_token);
+
+    let user = match creds.fetch_user().await {
+        Ok(user) => user,
+        Err(err) => {
+            tracing::error!("{err:?}");
+            return;
+        }
+    };
+    println!("Logged as {} ({})", user.name, user.email);
+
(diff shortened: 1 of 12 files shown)


crates/cli/src/commands/auth/whoami.rs
```diff
@@ -0,0 +1,17 @@
...
+
+    let creds = Credentials::load().unwrap();
+    if creds.api_token.is_none() {
+        eprintln!("Not logged in");
+        return;
+    }
+
+    let user = creds.fetch_user().await.unwrap();
+
+    println!("Logged in as:");
```
alexcasalboni, line 14: Maybe this could log the user ID as well, for debugging?

alexcasalboni:
I understand the reasons for using sub-modules to structure our internal code/folders, but do we really need two layers/hierarchies of commands?

This isn't bad:

```bash 
$ edgee auth login
```

But imho this is much simpler and easier to remember:

```
$ edgee login
```

crates/cli/src/commands/auth/login.rs
```diff
@@ -0,0 +1,39 @@
...
+    let confirm_overwrite =
+        Confirm::new("An API token is already present, do you want to overwrite it?")
+            .with_default(false);
+    if creds.api_token.is_some() && !confirm_overwrite.prompt().unwrap() {
+        return;
+    }
+
+    let api_token = Password::new("Enter Edgee API token (press Ctrl+R to toggle input display):")
+        .with_help_message(
+            "You can create one at https://www.edgee.cloud/<username>/settings/tokens",
```
CLEMENTINATOR, line 19: i wonder if theres a deep link (one that work for every user without needing the username) to bring you to the correct page

crates/components/src/auth/mod.rs
```diff
@@ -0,0 +1,86 @@
...
+    pub async fn fetch_user(&self) -> Result<User> {
+        use reqwest::Client;
+
+        let Some(ref api_token) = self.api_token else {
+            anyhow::bail!("No API token provided");
+        };
+
+        let client = Client::new();
+        let res = client
+            .get("https://api.edgee.app/v1/users/me")
```
CLEMENTINATOR, line 77: maybe this should be taken from the config ? so we can use it in our dev envs without modifying the url in the code

CLEMENTINATOR (commented):
small feedback

crates/cli/src/commands/auth/login.rs
CLEMENTINATOR: lets go with it ! :rocket: so the user can just click !

crates/components/src/auth/mod.rs
KokaKiwi: It doesn't really make sense to me to put that in the config since that's a dev-only feature (as in: not something an end-user would have to know)
Also the `edgee-components` crate does not have access to the server config so it'd be kind of tricky

Instead it can be configured by an environment var `EDGEE_API_ENDPOINT_BASE_URL` (which defaults to `https://api.edgee.cloud`) which can be overriden on a dev env (using dotenv, direnv, mise, etc.)

CLEMENTINATOR (approved):
will test tommorow but lgtm !

alexcasalboni (approved):
🚀

PR #150, opened 2024-12-14
Refactor/create protocols wit artifact

flyaruu:
### Checklist

See linear issue EOSS-46 (can I move that to github somehow?)

I've created another repo, called edgee-wit which hosts the wit file (and possibly dependencies if we would ever need them), and include them based on the url to the release.

We still need to think about verification and automation on that side, it is manual for now, I'll explain the steps in its readme.

To edit dependencies edit the crates/wasmtime/wit/deps.toml
...
and run wit-deps in the crates/wasmtime folder.

### Description of Changes

Describe your changes here

### Related Issues

List related issues here


crates/wasmtime/wit/deps/protocols/protocols.wit
```diff
@@ -0,0 +1,113 @@
+package edgee:protocols;
```
alexcasalboni, line 1: should this file be .gitignored? or it still makes sense to keep a local copy in each repository that has it as a dependency?

CLEMENTINATOR: ive asked myself the same thing while looking at it this morning

alexcasalboni: I don't have a super strong opinion, but I'm inclined to .gitignore it, as I'd do for any other external dependency (to provide a best practice for component developers, to avoid the temptation of updating it manually to make the CI/CD work, etc.)

flyaruu: A few days ago I would have agreed, but after a few days of WIT wrangling I lean the other way. If I'd update a dependency and the Rust code starts breaking in random places, the very first thing I'd like to know is what changed exactly in the wit tree. Maybe revert a particular change to the WIT to see the effect is. We can figure that out by downloading the old and new archives, unpacking and diffing them, but that information we would get for free from git.

flyaruu: So if there are no strong opinions, I'd like to keep them for the time being

alexcasalboni: cool 👍  and then each component developer can decide what to do in their repo anyway

SachaMorard (approved):
I approved!
But I have an open question: Is it really more practical to use wit-deps than a simple crates lib?
I understand reading the code and edgee-wit repo that a`wit-deps` command downloads and write the wit file in the directory...

alexcasalboni:
> But I have an open question: Is it really more practical to use wit-deps than a simple crates lib?

I think it makes sense especially for OSS developers that will develop components in other programming languages (other than Rust).

flyaruu:
It also makes it easier for us to use existing standards. 

For example, there is a draft WASI standard for logging inside WASM: https://github.com/WebAssembly/wasi-logging/tree/main

If we supply this from the host, any component gets 'free' standard logging support, and can just import that.

So yes, I think it is worth it.

SachaMorard:
> > But I have an open question: Is it really more practical to use wit-deps than a simple crates lib?
> 
> I think it makes sense especially for OSS developers that will develop components in other programming languages (other than Rust).

`wit-deps` is a rust crate, not sure a JS or Python developer will use it easily.
And it's not a standard either (only few downloads).
But we'll see how useful and practical this is.

PR #156, opened 2024-12-20
Created WasmSource enum to support resolving Wasm code from different sources + allow dynamic configuration

flyaruu:
…y to create other (dynamic) ways of configuring

### Checklist

* [v] I have read the [Contributor Guide](../../CONTRIBUTING.md)
* [v] I have read and agree to the [Code of Conduct](../../CODE_OF_CONDUCT.md)
* [v] I have added a description of my changes and why I'd like them included in the section below

### Description of Changes

Refactored the concrete types of the configuration into traits, and made the original types into implementations of that.
This will allow us to create other (non file based) implementations for different situations.

### Related Issues

List related issues here

### DIFF
--- a/crates/cli/src/commands/serve.rs
+++ b/crates/cli/src/commands/serve.rs
@@ -1,7 +1,7 @@
 setup_command! {}
 
 pub async fn run(_opts: Options) {
-    edgee_server::init().unwrap();
+    edgee_server::init().await.unwrap();
 
     tokio::select! {
         Err(err) = edgee_server::monitor::start() => tracing::error!(?err, "Monitor failed"),
--- a/crates/server/src/config.rs
+++ b/crates/server/src/config.rs
@@ -1,7 +1,7 @@
 use std::collections::HashSet;
 use std::sync::OnceLock;
 
-use edgee_wasmtime::components::config::ComponentsConfiguration;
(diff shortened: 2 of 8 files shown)


KokaKiwi:
I don't think we need to replace the whole configuration type to a provider-type trait

If we need another way to provide the component content, it could be simply replaced by changing the type to a `Vec<u8>` (instead of a path) containing directly the component content (or a `Rc<[u8]>` if we need a thread-safe way to share it but it should not be needed as we should not need to clone the struct) or to a dynamic trait-type for providing lazily the content (for example something like `Box<dyn Fn() -> Vec<u8>>` or something like that)

Rust traits can be really tricky to use and can burden the code readability a lot, so they need to be used sparingly and only when/where needed :)

flyaruu:
My goal here is mostly to reduce the coupling between the way of configuration and the configuration itself.

The original concrete types map to the format of the edgee.toml file and assume the wasm component is on a filesystem and reachable by a a certain path.

From the wasmtime component none of that matters, it needs a list of components, each capable of supplying a hashmap of credentials and the binary wasm data.

So that is what the configuration api should accept, I think a trait is the best tool for that. The Rust code I tend to see prefers static dispatch traits over Box<dyn> traits so I went with that, but I don't have a strong opinion there.

Would a dynamic trait work for you?

PR #157, opened 2024-12-23
feat: add sec-ch-ua headers to analytics requests

SachaMorard:
### Checklist

### Description of Changes

To replicate browser behavior, we add sec-ch-ua headers to data collection requests

### Related Issues

No issue


crates/wasmtime/src/components/mod.rs
```diff
@@ -316,6 +316,100 @@ fn insert_expected_headers(
         );
     }
 
+    // Insert sec-ch-ua headers
+    // sec-ch-ua
+    if event
```
CLEMENTINATOR, line 321: maybe im wrong, but using a if let(Some(xxx))) would allow you to remove the same big pattern below

CLEMENTINATOR:
```
if let Some(ua_version_list) =  .context
        .as_ref()
        .unwrap()
        .client
        .as_ref()
        .unwrap()
        .user_agent_version_list {
                headers.insert(
            HeaderName::from_str("sec-ch-ua")?,
            HeaderValue::from_str(
                format_ch_ua_header(ua_version_list)

        }

CLEMENTINATOR (approved):
other than my comment, lgtm !

crates/wasmtime/src/components/mod.rs
SachaMorard: You're right. I made it simpler ;)

CLEMENTINATOR: TIL we can even do it better by using https://doc.rust-lang.org/std/option/enum.Option.html#method.and_then instead of nested ifs, but your way also works !

PR #168, opened 2025-01-09
fix: cookie parsing is altered when a cookie contains an equal char

SachaMorard:
### Checklist

### Description of Changes

When Edgee parses all the cookies, a cookie that contains an equal sign disappears from the cookies list

### Related Issues

No related issue

### DIFF
--- a/crates/server/src/proxy/context/proxy.rs
+++ b/crates/server/src/proxy/context/proxy.rs
@@ -64,16 +64,13 @@ impl<'a> ProxyContext<'a> {
             let cookie_str = cookie.to_str().unwrap();
             let cookie_parts: Vec<&str> = cookie_str.split(";").map(|c| c.trim()).collect();
             for cookie_part in cookie_parts {
-                let parts: Vec<&str> = cookie_part.split('=').collect();
-                if parts.len() != 2 {
-                    continue;
-                }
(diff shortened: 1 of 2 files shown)


CLEMENTINATOR (approved):
lgtm. I dont know if we have any but a testsuite on these functions could be interesting

crates/server/src/proxy/context/proxy.rs
```diff
@@ -65,7 +65,7 @@ impl<'a> ProxyContext<'a> {
             let cookie_parts: Vec<&str> = cookie_str.split(";").map(|c| c.trim()).collect();
             for cookie_part in cookie_parts {
                 let parts: Vec<&str> = cookie_part.split('=').collect();
-                if parts.len() != 2 {
+                if parts.len() < 2 {
                     continue;
                 }
```
KokaKiwi, line 70: As stated elsewhere, I would suggest using [`str::split_once`](https://doc.rust-lang.org/std/primitive.str.html#method.split_once) instead, which is more fitting for this use-case

```rust
let Some((name, value)) = cookie.split_once('=') else {
  continue;
};
```

SachaMorard: @KokaKiwi can you check again?

KokaKiwi (approved):

PR #184, opened 2025-01-23
chore: component runtime refactoring

SachaMorard:
### Checklist

### Description of Changes

New organization of components-runtime crate to be able to welcome new categories of components in the future

#### World
Each category of components will have its dedicated world 
```wit
package edgee:native;

world data-collection {
  export edgee:protocols/data-collection;
}

world consent-mapping {
  export edgee:protocols/consent-mapping;
}
``` 

I thought it would be possible to create a single `edgee` world, which would export several interfaces (one for `data-collection`, another for `consent-mapping`...), and that a component could use only one interface. However, it seems that this is not possible. 
If a component uses only one interface, the compilation works, but using it in `edgee-components-runtime` fails. 

Consequently, a single world must be created for each component category. 

### New wit dependencies

Wit deps is now 
```
protocols="https://github.com/edgee-cloud/edgee-wit/archive/refs/tags/v0.3.0.tar.gz"
```

### Data Collection Payload

Data Collection payload is now handled by the proxy server. But components-runtime has its own Data Collection Payload struct to simplify the portability of the data, and to make components-runtime more autonomous

### Component context

Component context is now able to handle different categories of components. 
For now, only the data collection components are operational, but I already prepared the context to instanciate consent_mapping components

### Related Issues

No issue


CLEMENTINATOR (approved):
looks pretty good !

crates/components-runtime/src/config.rs
```diff
@@ -0,0 +1,80 @@
+use std::{collections::HashMap, path::PathBuf};
+
+use serde::Deserialize;
+
+#[derive(Deserialize, Debug, Default, Clone)]
+pub struct ComponentsConfiguration {
+    #[serde(default = "default_data_collection_components")]
+    pub data_collection: Vec<DataCollectionComponents>,
+    #[serde(default = "default_consent_mapping_components")]
+    pub consent_mapping: Vec<ConsentMappingComponents>,
```
KokaKiwi, line 10: On this case, since the default is an empty `Vec`, you can simply use `#[serde(default)]` here so it will use the `std::default::Default` trait which is an empty vec for this type :)

```suggestion
    #[serde(default)]
    pub data_collection: Vec<DataCollectionComponents>,
    #[serde(default)]
    pub consent_mapping: Vec<ConsentMappingComponents>,
```

crates/components-runtime/src/context.rs
```diff
@@ -32,41 +36,85 @@ impl ComponentsContext {
...
+        // Data collection components
+        let data_collection_components = config
             .data_collection
             .iter()
             .map(|entry| {
                 let span = tracing::info_span!("component-context", component = %entry.name);
                 let _span = span.enter();
 
-                tracing::debug!("Start pre-instanciate component");
+                tracing::debug!("Start pre-instantiate data collection component");
```
alexcasalboni, line 47: is it worth logging which component here?

crates/components-runtime/src/config.rs
```diff
@@ -0,0 +1,80 @@
...
+    vec![]
+}
+
+#[derive(Deserialize, Debug, Default, Clone)]
+pub struct DataCollectionComponents {
+    pub name: String,
+    pub component: String,
+    pub credentials: HashMap<String, String>,
+    #[serde(default = "default_data_collection_component_config")]
+    pub config: DataCollectionComponentConfig,
```
KokaKiwi, line 28: `Default` is already implemented on `DataCollectionComponentConfig` so no need for special functon

```suggestion
    #[serde(default)]
    pub config: DataCollectionComponentConfig,
```

crates/components-runtime/src/config.rs
```diff
@@ -0,0 +1,80 @@
...
+    pub anonymization: bool,
+    #[serde(default = "default_pending")]
+    pub default_consent: String,
+    #[serde(default = "default_true")]
+    pub track_event_enabled: bool,
+    #[serde(default = "default_true")]
+    pub user_event_enabled: bool,
+    #[serde(default = "default_true")]
+    pub page_event_enabled: bool,
+}
```
KokaKiwi, line 43: I think you can replace the field-specific serde default attributes by a struct-global serde default attribute instead
It would use the `Default` trait of the struct (https://serde.rs/container-attrs.html#default )

```suggestion
#[serde(default)]
pub struct DataCollectionComponentConfig {
    pub anonymization: bool,
    pub default_consent: String,
    pub track_event_enabled: bool,
    pub user_event_enabled: bool,
    pub page_event_enabled: bool,
}
```

crates/components-runtime/src/config.rs
```diff
@@ -0,0 +1,80 @@
...
+
+fn default_pending() -> String {
+    "pending".to_string()
+}
+
+#[derive(Deserialize, Debug, Default, Clone)]
+pub struct ConsentMappingComponents {
+    pub name: String,
+    pub component: String,
+    #[serde(default = "default_consent_mapping_component_config")]
```
KokaKiwi, line 67: Same as above, let's use the `Default` trait whenever possible (helps with readability and maintenance)

crates/components-runtime/src/context.rs
```diff
@@ -32,41 +36,85 @@ impl ComponentsContext {
         let mut linker = Linker::new(&engine);
         wasmtime_wasi::add_to_linker_async(&mut linker)?;
 
-        let components = config
+        // Data collection components
+        let data_collection_components = config
             .data_collection
             .iter()
             .map(|entry| {
                 let span = tracing::info_span!("component-context", component = %entry.name);
```
KokaKiwi, line 44: Maybe add the component category here as a span attribute instead of in the trace messages?

```suggestion
                let span = tracing::info_span!("component-context", component = %entry.name, category = "data-collection");
```

alexcasalboni (approved):
LGTM 👍

SachaMorard:
@KokaKiwi I made the changes you suggested. Can you check?

crates/components-runtime/src/context.rs
```diff
@@ -0,0 +1,149 @@
...
+
+        // Data collection components
+        let data_collection_components = config
+            .data_collection
+            .iter()
+            .map(|entry| {
+                let span = tracing::info_span!("component-context", component = %entry.name, category = "data-collection");
+                let _span = span.enter();
+
+                tracing::debug!("Start pre-instantiate data collection component");
```
KokaKiwi, line 47: We don't need to specify the "data collection" here now that's already attached to the span :)

KokaKiwi (approved):

crates/components-runtime/src/context.rs
SachaMorard: That was your suggestion ;)

KokaKiwi: I meant that the "data collection" part was already in the span, so the message does not need to mention it again ^^
It would output something like this:
`DEBUG {component="...", category="data-collection"} Start pre-instantiate data collection component`

PR #185, opened 2025-01-23
feat: debugger api

SachaMorard:
### Checklist

### Description of Changes

In order for an Edgee user to use the console to debug component events, Components-runtime must use the edgee API to send details of the requests and responses made.

By default, no call is made to the edgee API. 

### Related Issues

No issue


crates/components-runtime/src/data_collection/debug.rs
```diff
@@ -0,0 +1,290 @@
...
+        let content_type = edgee_request
+            .headers
+            .iter()
+            .find(|(k, _)| k.to_lowercase() == "content-type")
+            .map(|(_, v)| v.to_string())
+            .unwrap_or_else(|| "text/plain".to_string());
+
+        let b: Option<serde_json::Value> = if edgee_request.body.is_empty() {
+            None
+        } else if content_type.contains("application/json") || component_slug == "piano_analytics" {
```
alexcasalboni, line 113: why an hard-coded component slug? isn't the content type application/json for piano?

crates/components-runtime/src/data_collection/mod.rs
```diff
@@ -27,27 +29,40 @@ pub async fn send_json_events(
...
 
     let mut events: Vec<Event> = serde_json::from_str(events_json)?;
-    send_events(ctx, &mut events, component_config, log_component).await
+    send_events(
+        ctx,
+        &mut events,
+        component_config,
+        trace_component,
+        debug,
+        "",
```
alexcasalboni, line 46: is project_id always meant to be empty here?

SachaMorard: Yes, because this function is always called by edgee OSS to use components locally, and so, there is no "project_id" or "proxy_host"

crates/components-runtime/src/data_collection/debug.rs
SachaMorard: You're right @alexcasalboni 
It's fixed

crates/components-runtime/src/data_collection/logger.rs
```diff
@@ -0,0 +1,61 @@
...
+            $component_name,
+            $status,
+            $event_type,
+            $path,
+            $consent,
+            $duration,
+            $message,
+        );
+    }};
+}
```
KokaKiwi, line 61: Why make a macro instead of directly calling the function (since that's what's this macro actually does)?

alexcasalboni (approved):

crates/components-runtime/src/data_collection/logger.rs
SachaMorard: Right! And we don't need this script for now. I removed it

PR #191, opened 2025-01-27
feat: add redirection basic logic

NicolasGirardot:
### Checklist

### Description of Changes

In this Pull Request, we add the capacity for redirections. For now they are quite basic but can be configured through the TOML file 

```toml
# edgee.toml
[[redirections]]
origin = "http://example.com/old-path"
destination = "http://example.com/new-path"
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,8 +187,8 @@ Here's how you can set up a redirection in your `edgee.toml` configuration file:
 ```toml
 # edgee.toml
 [[redirections]]
-origin = "http://example.com/old-path"
-destination = "http://example.com/new-path"
+source = "http://example.com/old-path"
+target = "http://example.com/new-path"
 ```
 
 In this example, requests to `http://example.com/old-path` will be temporarily (HTTP 302) redirected to `http://example.com/new-path`
--- a/README.md
+++ b/README.md
@@ -191,7 +191,7 @@ origin = "http://example.com/old-path"
 destination = "http://example.com/new-path"
 ```
 
-In this example, requests to `http://example.com/old-path` will be temporarily redirected to `http://example.com/new-path`
+In this example, requests to `http://example.com/old-path` will be temporarily (HTTP 302) redirected to `http://example.com/new-path`
 
 
 ## Integrating with edgee components
--- a/README.md
+++ b/README.md
@@ -176,6 +176,24 @@ backend = "api"
 
 This way, calling `/api/test` on port `80` will result in calling `/v1/test` on the API backend.
 
+## Redirections
+
+Edgee supports HTTP redirections, allowing you to redirect traffic from one URL to another. This can be useful for various purposes, such as redirecting HTTP to HTTPS, or setting up permanent or temporary redirects.
+
+### Example
+
+Here's how you can set up a redirection in your `edgee.toml` configuration file:
+
+```toml
+# edgee.toml
+[[redirections]]
+origin = "http://example.com/old-path"
+destination = "http://example.com/new-path"
+```
+
+In this example, requests to `http://example.com/old-path` will be temporarily redirected to `http://example.com/new-path`
+
+
 ## Integrating with edgee components
 
 Check out the [official components docs](https://www.edgee.cloud/docs/components/overview) to dive into the
--- a/README.md
+++ b/README.md
@@ -176,22 +176,28 @@ backend = "api"
 
 This way, calling `/api/test` on port `80` will result in calling `/v1/test` on the API backend.
 
-## Redirections
+### Redirections
 
-Edgee supports HTTP redirections, allowing you to redirect traffic from one URL to another. This can be useful for various purposes, such as redirecting HTTP to HTTPS, or setting up permanent or temporary redirects.
+Edgee supports HTTP redirections, allowing you to redirect traffic temporarily from one URL to another. 
 
-### Example
+#### Example
 
 Here's how you can set up a redirection in your `edgee.toml` configuration file:
 
 ```toml
 # edgee.toml
-[[redirections]]
-source = "http://example.com/old-path"
-target = "http://example.com/new-path"
+[[routing.redirections]]
+source = "/old-path"
+target = "https://example.com/new-path"
+
+
+[[routing.redirections]]
+source = "/foo"
+target = "/bar"
 ```
 
-In this example, requests to `http://example.com/old-path` will be temporarily (HTTP 302) redirected to `http://example.com/new-path`
+In this example, requests to `https://demo.edgee.dev/old-path` will be temporarily (HTTP 302) redirected to `https://example.com/new-path` and requests to `https://demo.edgee.dev/foo` will be redirected to `https://demo.edgee.dev/bar`
+
 
 
 ## Integrating with edgee components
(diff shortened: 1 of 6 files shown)


crates/server/src/proxy/context/redirection.rs
```diff
@@ -0,0 +1,30 @@
...
+                    == format!(
+                        "{}://{}{}",
+                        request.get_proto(),
+                        request.get_host(),
+                        request.get_path()
+                    )
+            })?
+            .to_owned();
+
+        Some(Self {
```
CLEMENTINATOR, line 25: could chain the find with a map 

```
        .map(|redirection| Self {
            _origin: redirection.origin.clone(),
            destination: redirection.destination.clone(),
        })
```

crates/server/src/proxy/context/redirection.rs
```diff
@@ -0,0 +1,30 @@
+use super::incoming::RequestHandle;
+use crate::config;
+
+pub struct RedirectionContext {
+    pub _origin: String,
```
CLEMENTINATOR, line 5: why _ ?

README.md
```diff
@@ -176,6 +176,24 @@ backend = "api"
...
+Here's how you can set up a redirection in your `edgee.toml` configuration file:
+
+```toml
+# edgee.toml
+[[redirections]]
+origin = "http://example.com/old-path"
+destination = "http://example.com/new-path"
+```
+
+In this example, requests to `http://example.com/old-path` will be temporarily redirected to `http://example.com/new-path`
```
alexcasalboni, line 194: I would clarify what we mean by "temporarily" (does it mean 302 status code?)

crates/server/src/config.rs
```diff
@@ -78,6 +81,23 @@ impl StaticConfiguration {
...
+        let mut duplicates = HashSet::new();
+
+        for redirection in &self.redirections {
+            if !seen.insert(&redirection.origin) {
+                duplicates.insert(&redirection.origin);
+            }
+        }
+
+        if !duplicates.is_empty() {
+            Err(format!("duplicate redirections found: {:?}", duplicates))
```
alexcasalboni, line 96: should this be just a warning? I understand the logic, but it is worth breaking everything because of a duplicated redirect?

crates/server/src/proxy/context/redirection.rs
CLEMENTINATOR: function could then be rewritten as 

```
    pub fn from_request(request: &RequestHandle) -> Option<Self> {
        let cfg = &config::get().redirections;
        cfg
            .iter()
            .find(|r| {
                r.origin.as_str()
                    == format!(
                        "{}://{}{}",
                        request.get_proto(),
                        request.get_host(),
                        request.get_path()
                    )
            })
            .map(|redirection| Self {
            _origin: redirection.origin.clone(),
            destination: redirection.destination.clone(),
        })
   }
```

crates/server/src/proxy/controller.rs
```diff
@@ -185,6 +186,18 @@ pub fn redirect_to_https(request: &RequestHandle) -> anyhow::Result<Response> {
         .expect("response builder should never fail"))
 }
 
+pub fn build_redirection(associated_redirection: &RedirectionContext) -> anyhow::Result<Response> {
+    Ok(http::Response::builder()
+        .status(StatusCode::MOVED_PERMANENTLY)
```
alexcasalboni, line 191: so it's 301 by default? until we make this configurable, I'd suggest we use 302 by default because 301 is cached by browsers and it's hard to revert back

CLEMENTINATOR: :+1:

crates/server/src/proxy/context/redirection.rs
```diff
@@ -0,0 +1,30 @@
...
+        let cfg = &config::get().redirections;
+        let redirection = cfg
+            .iter()
+            .find(|r| {
+                r.origin.as_str()
+                    == format!(
+                        "{}://{}{}",
+                        request.get_proto(),
+                        request.get_host(),
+                        request.get_path()
```
SachaMorard, line 20: Is not better to look at `path` instead of the full URL (`proto` + `domain` + `path`)

SachaMorard (changes requested):

crates/server/src/proxy/context/redirection.rs
NicolasGirardot: Not used for now but could be later ? not so sure

CLEMENTINATOR: ok !

CLEMENTINATOR: imo just remove it !

README.md
```diff
@@ -176,6 +176,24 @@ backend = "api"
...
+
+### Example
+
+Here's how you can set up a redirection in your `edgee.toml` configuration file:
+
+```toml
+# edgee.toml
+[[redirections]]
+source = "http://example.com/old-path"
+target = "http://example.com/new-path"
```
SachaMorard, line 191: As redirection sources are based on paths, your example is not right

crates/server/src/proxy/mod.rs
```diff
@@ -116,6 +117,10 @@ pub async fn handle_request(
         return controller::empty_json_response();
     }
 
+    if let Some(redirection_ctx) = RedirectionContext::from_request(request) {
```
SachaMorard, line 120: To follow the log logic, you should add an info here:
```rust
info!(
    "302 - {} {}{} - {}ms",
    request.get_method(),
    request.get_host(),
    request.get_path(),
    timer_start.elapsed().as_millis()
);
```

crates/server/src/config.rs
```diff
@@ -25,6 +26,9 @@ pub struct StaticConfiguration {
 
     #[serde(default)]
     pub components: ComponentsConfiguration,
+
+    #[serde(default)]
+    pub redirections: Vec<RedirectionsConfiguration>,
```
SachaMorard, line 31: Would it be better to add RedirectionsConfiguration in RoutingConfiguration?

The way you build it means a user can create global redirection rules. But `edgee` can listen to multiple domains.
Let's say I configured `edgee` to serve the `demo.edgee.dev` and `test.edgee.dev` domains, each with its routing configuration.

Then I create the following redirection:
```rust
[[redirections]]
source = "/old-path"
target = "/new-path"
```
The redirection will be enabled to the two domains
`demo.edgee.dev/old-path` -> `demo.edgee.dev/new-path`
`test.edgee.dev/old-path` -> `test.edgee.dev/new-path`

Would it be better to do something like:
```rust
[[routing]]
domain = "demo.edgee.dev"

[[routing.redirections]]
source = "/old-path"
target = "/new-path"

[[routing.backends]]
name = "demo"
default = true
address = "localhost:8080"
enable_ssl = false
```

SachaMorard (changes requested):

CLEMENTINATOR (approved):

alexcasalboni (approved):

SachaMorard (approved):

PR #193, opened 2025-01-28
feat: use env var to test api locally

SachaMorard:
### Checklist

### Description of Changes

To test edgee with a local version of the api, two env vars can be used to change the API base url and the API token

### Related Issues

No issue

### DIFF
--- a/crates/api-client/src/lib.rs
+++ b/crates/api-client/src/lib.rs
@@ -22,6 +22,12 @@ pub fn connect(#[builder(default = PROD_BASEURL)] baseurl: &str, api_token: Stri
 
     let mut default_headers = HeaderMap::new();
 
+    // if API_URL env var is set, redefine baseurl
+    let baseurl = std::env::var("API_URL").unwrap_or(baseurl.to_string());
+
+    // if API_TOKEN env var is set, redefine api_token
+    let api_token = std::env::var("API_TOKEN").unwrap_or(api_token.to_string());
+
     let auth_value = format!("Bearer {api_token}");
     default_headers.insert(header::AUTHORIZATION, auth_value.try_into().unwrap());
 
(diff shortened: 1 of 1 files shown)


CLEMENTINATOR (approved):
weird, I thought @KokaKiwi did that already 🤔 lgtm !

crates/api-client/src/lib.rs
```diff
@@ -22,6 +22,12 @@ pub fn connect(#[builder(default = PROD_BASEURL)] baseurl: &str, api_token: Stri
 
     let mut default_headers = HeaderMap::new();
 
+    // if API_URL env var is set, redefine baseurl
+    let baseurl = std::env::var("API_URL").unwrap_or(baseurl.to_string());
```
alexcasalboni, line 26: maybe `EDGEE_API_URL` and `EDGEE_API_TOKEN` would be more specific?

KokaKiwi (changes requested):
There's a small problem there, if you pass a custom base URL but the env var is present, the env var will overwrite what the crate's user passed to the function, which seems to be a big side-effect to me

For the context, I intended for this function to be used by the crate's user to specify a custom URL if needed (from an env var or something else), so the function was meant to be the most simple since passing custom params (for local dev for example) was already handled

I would suggest instead to create a separate function `connect_from_env` which would basically do the same as here (with `PROD_BASEURL` as the default for `EDGEE_API_URL`) and call the `connect` function with that

alexcasalboni:
Agree 👍  My suggestion would be to use the env variable as a fallback/default in case no base URL is passed (instead of using the param as fallback).

SachaMorard:
> With that said maybe not that much of a side-effect actually, since the edgee CLI config is also fetched in a way the env is prioritized over the config file (which is similar in behaviour as here)

@KokaKiwi I'm not shocked when an environment variable overwrites parameters. This is deliberate, because in production, we need to use environment variables rather than long command lines with parameters all over the place.

KokaKiwi:
> I'm not shocked when an environment variable overwrites parameters. This is deliberate, because in production, we need to use environment variables rather than long command lines with parameters all over the place.

Yep that's why I put an "Edit", I just realized it was an OK behaviour actually :sweat_smile:

PR #194, opened 2025-01-28
feat: data collection debug endpoints change

SachaMorard:
### Checklist

### Description of Changes

To reflect API changes: 
- data collection events endpoints now have a `/debug` prefix
- we now use an `Authorization` header instead of `x-api-secret` when debug info is sent to the API from data collection components

### Related Issues

List related issues here

### DIFF
--- a/crates/components-runtime/src/data_collection/debug.rs
+++ b/crates/components-runtime/src/data_collection/debug.rs
@@ -307,7 +307,7 @@ pub async fn debug_and_trace_response(
         let api_url = std::env::var("EDGEE_API_URL").unwrap_or_default();
 
         if !api_super_token.is_empty() && !api_url.is_empty() && !params.project_id.is_empty() {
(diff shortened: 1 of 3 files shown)


crates/components-runtime/src/data_collection/debug.rs
```diff
@@ -32,12 +32,16 @@ pub struct DebugPayload {
 }
 
 impl DebugPayload {
-    pub fn new(params: DebugParams) -> Self {
+    pub fn new(params: DebugParams, error: &str) -> Self {
         let component_request = DebugComponentRequest::new(&params.request);
 
+        let mut body = params.response_body;
+        if !error.is_empty() {
+            body = Some(error.to_string())
```
alexcasalboni, line 40: are we ok with completely discarding `params.response_body` in this case? anything useful there in addition to the error message?

alexcasalboni (approved):
LGTM 👍  Just a minor doubt

PR #196, opened 2025-01-30
add forward_header option to data collection component

CLEMENTINATOR:
### Checklist

### Description of Changes

add an option for data collection component to skip forwarding headers from the client to the data collection endpoint

### DIFF
--- a/crates/components-runtime/src/config.rs
+++ b/crates/components-runtime/src/config.rs
@@ -20,6 +20,7 @@ pub struct DataCollectionComponents {
     pub credentials: HashMap<String, String>,
     #[serde(default)]
     pub config: DataCollectionComponentConfig,
+    pub forward_headers: Option<bool>,
 }
 
 #[derive(Deserialize, Debug, Clone)]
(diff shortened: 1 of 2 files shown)


alexcasalboni (approved):
Wow 🚀 ❤️

crates/components-runtime/src/config.rs
```diff
@@ -20,6 +20,7 @@ pub struct DataCollectionComponents {
     pub credentials: HashMap<String, String>,
     #[serde(default)]
     pub config: DataCollectionComponentConfig,
+    pub forward_headers: Option<bool>,
```
SachaMorard, line 23: I'm wondering about the name of this parameter. 
Would it be better to use `extra_headers` or `enrich_headers`?

SachaMorard:
@clement and you'll have to add this to data-collection as well

crates/components-runtime/src/config.rs
CLEMENTINATOR: `forward_client_headers` maybe ?

SachaMorard: Perfect

PR #197, opened 2025-01-31
feat: Add components subcommand

KokaKiwi:
### Checklist

### Description of Changes

- Add the `edgee components` subcommand
- Some refactor to adapt the CLI to subcommands other than `edgee serve`

### Related Issues

- EDGEE-240


crates/cli/src/commands/components/push.rs
```diff
@@ -0,0 +1,102 @@
...
+    tracing::info!("Uploading output artifact...");
+    let asset_url = client
+        .upload_file(&manifest.package.build.output_path)
+        .await
+        .expect("Could not upload component");
+
+    tracing::info!("Creating component version...");
+    client
+        .create_component_version()
+        .id(&manifest.package.name)
```
CLEMENTINATOR, line 84: this will need to be changed now that we have the new slug definition

crates/cli/src/commands/components/push.rs
```diff
@@ -0,0 +1,102 @@
...
+        .body(
+            types::ComponentVersionCreateInput::builder()
```
CLEMENTINATOR, line 86: mssing "configuration" :(

crates/cli/src/commands/components/build.rs
```diff
@@ -0,0 +1,26 @@
+#[derive(Debug, clap::Parser)]
+pub struct Options {}
+
+pub async fn run(_opts: Options) -> anyhow::Result<()> {
```
alexcasalboni, line 4: we're not using the `output_path` from manifest at all here? so we assume the build command will output the file in the correct place, right?

I remember we said we wanted to implement a default command for each language (so it doesn't have to be configured explicitly) with the ability to overwrite/customize it in the manifest

crates/cli/src/commands/components/build.rs
```diff
@@ -0,0 +1,26 @@
...
+
+    let Some(manifest_path) = manifest::find_manifest_path() else {
+        anyhow::bail!("Manifest not found");
+    };
+    let manifest = Manifest::load(&manifest_path).map_err(|err| anyhow::anyhow!(err))?;
+
+    tracing::info!("Running: {}", manifest.package.build.command);
+    let mut cmd = Command::new("sh");
+    cmd.arg("-c").arg(manifest.package.build.command);
+    let status = cmd.status()?;
```
alexcasalboni, line 17: I see that:

> By default, stdin, stdout and stderr are inherited from the parent.

Is that ok by default? maybe it shouldn't be logged to stdout unless you set an env variable for debugging?

crates/cli/src/commands/components/build.rs
```diff
@@ -0,0 +1,26 @@
+#[derive(Debug, clap::Parser)]
+pub struct Options {}
+
+pub async fn run(_opts: Options) -> anyhow::Result<()> {
+    use std::process::Command;
+
+    use crate::components::manifest::{self, Manifest};
+
+    let Some(manifest_path) = manifest::find_manifest_path() else {
+        anyhow::bail!("Manifest not found");
```
alexcasalboni, line 10: (see the error message in the `check` command, something like "Edgee Manifest not found. Please run `edgee component create` and start from a template or `edgee component init` to create a new empty manifest in this folder.")

crates/cli/src/commands/components/push.rs
```diff
@@ -0,0 +1,102 @@
...
+pub async fn run(opts: Options) -> anyhow::Result<()> {
+    use edgee_api_client::{auth::Credentials, types, ResultExt};
+
+    use crate::components::manifest::{self, Manifest};
+
+    let creds = Credentials::load()?;
+    creds.check_api_token()?;
+
+    let Some(manifest_path) = manifest::find_manifest_path() else {
+        anyhow::bail!("Manifest not found");
```
alexcasalboni, line 18: same (see `check` error message)

crates/cli/src/commands/components/push.rs
```diff
@@ -0,0 +1,102 @@
...
+    };
+    let manifest = Manifest::load(&manifest_path)?;
+
+    let client = edgee_api_client::new()
+        .baseurl("https://api.edgee.dev")
```
alexcasalboni, line 23: should this be the default base url? (so you don't have to set it every time & everywhere)

crates/cli/src/commands/components/push.rs
```diff
@@ -0,0 +1,102 @@
...
+        .credentials(&creds)
+        .connect();
+
+    let organization = match opts.organization {
+        Some(ref organization) => client
+            .get_organization()
+            .id(organization)
+            .send()
+            .await
+            .api_with_context(|| format!("Could not get organization `{organization}`"))?
```
alexcasalboni, line 33: this could also explode if you aren't part of that organization - can we try to log the error message and/or status code?

crates/cli/src/commands/components/push.rs
```diff
@@ -0,0 +1,102 @@
...
+            .get_my_organization()
+            .send()
+            .await
+            .api_context("Could not get user organization")?
+            .into_inner(),
+    };
+
+    match client
+        .get_component()
+        .id(&manifest.package.name)
```
alexcasalboni, line 45: do we already validate `manifest.package.name` to make sure it's a valid id/slug? (no spaces, no `/`, etc.)

crates/cli/src/commands/components/push.rs
```diff
@@ -0,0 +1,102 @@
...
+                        .repo_link(manifest.package.repository.map(|url| url.to_string())),
+                )
+                .send()
+                .await
+                .api_context("Could not create component")?;
+        }
+        Ok(_) | Err(_) => {}
+    }
+
+    tracing::info!("Uploading output artifact...");
```
alexcasalboni, line 75: I'd say "Uploading .wasm file..."

crates/cli/src/commands/components/push.rs
```diff
@@ -0,0 +1,102 @@
...
+        .await
+        .expect("Could not upload component");
+
+    tracing::info!("Creating component version...");
+    client
+        .create_component_version()
+        .id(&manifest.package.name)
+        .body(
+            types::ComponentVersionCreateInput::builder()
+                .version(&manifest.package.version)
```
alexcasalboni, line 87: what happens if this version already exists?

crates/cli/src/commands/components/push.rs
```diff
@@ -0,0 +1,102 @@
...
+                .wit_world_version(&manifest.package.wit_world_version)
+                .wasm_url(asset_url),
+        )
+        .send()
+        .await
+        .api_context("Could not create version")?;
```
alexcasalboni, line 93: can we include the API error message here? or is it automatically included via `api_context()`?

crates/cli/src/commands/components/test.rs
```diff
@@ -0,0 +1,173 @@
...
...":"https://demo.edgee.app/analytics-with-edgee.html","path":"/analytics-with-edgee.html","referrer":"https://demo.edgee.dev/analytics-with-js.html"},"context":{"page":{"keywords":["demo","tag manager"],"title":"Page with Edgee components","url":"https://demo.edgee.app/analytics-with-edgee.html","path":"/analytics-with-edgee.html","referrer":"https://demo.edgee.dev/analytics-with-js.html"},"user":{"edgee_id":"6bb171d5-2284-41ee-9889-91af03b71dc5"},"client":{"ip":"127.0.0.1","locale":"en-us","accept_language":"en-US,en;q=0.9","timezone":"Europe/Paris","user_agent":"Mozilla/5.0 (X11; Linux x86_64)AppleWebKit/537.36 (KHTML, like Gecko) Chrome/132.0.0.0 Safari/537.36","user_agent_version_list":"Not A(Brand;8|Chromium;132","user_agent_mobile":"0","os_name":"Linux","user_agent_architecture":"x86","user_agent_bitness":"64","user_agent_full_version_list":"Not A(Brand;8.0.0.0|Chromium;132.0.6834.159","user_agent_model":"","os_version":"6.12.11","screen_width":1920,"screen_height":1280,"screen_density":1.5},"session":{"session_id":"1738597536","session_count":1,"session_start":false,"first_seen":"2025-02-03T15:45:36.569004889Z","last_seen":"2025-02-03T15:46:39.278740029Z"}},"from":"edge"}]"#;
```
alexcasalboni, line 62: these are very useful, but ideally at some point we find a way to generate a local .json file and simply take a .json file reference as input parameter (similar to https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-cli-command-reference-sam-local-generate-event.html )

crates/cli/src/components/manifest.rs
```diff
@@ -0,0 +1,99 @@
...
+
+        let content = fs::read_to_string(path)
+            .with_context(|| format!("Could not read manifest file at {}", path.display()))?;
+
+        let manifest: Self = toml::from_str(&content)
+            .with_context(|| format!("Could not decode the manifest file at {}", path.display()))?;
+
+        if manifest.manifest_version != MANIFEST_VERSION {
+            anyhow::bail!(
+                "Invalid manifest version ({} != {})",
```
alexcasalboni, line 76: we could use a clearer message like "The only supported manifest version is XYZ"

crates/cli/src/commands/components/build.rs
KokaKiwi: The `output_path` field was meant for the `push` subcommand to find where the resulted build artifact was to be found, since we don't know (yet) where it's going
When we get the `language` field and pre-configured commands we'll get to that but in the meantime we still need this :)

crates/cli/src/components/manifest.rs
```diff
@@ -0,0 +1,99 @@
...
+    pub name: String,
+    pub version: String,
+    #[serde(with = "Category")]
+    pub category: api_types::ComponentCreateInputCategory,
+    #[serde(with = "SubCategory")]
+    pub subcategory: api_types::ComponentCreateInputSubcategory,
+    pub description: Option<String>,
+
+    #[serde(default)]
+    pub documentation: Option<url::Url>,
```
alexcasalboni, line 28: I would consider making `documentation` optional - and then if it's missing we simply use the `repository` value.

crates/cli/src/commands/components/push.rs
KokaKiwi: The `api_[with_]context` method actually add the error message returned from the API :)

crates/cli/src/commands/components/push.rs
KokaKiwi: That's on the API part, it rejects it and the CLI log the error saying "Version already exists" (the error message from the API)

crates/cli/src/commands/components/push.rs
KokaKiwi: Same as above

crates/cli/src/commands/components/test.rs
KokaKiwi: Poke @CLEMENTINATOR on this one :p

crates/cli/src/components/manifest.rs
```diff
@@ -0,0 +1,99 @@
...
+
+impl Manifest {
+    pub fn load(path: &Path) -> Result<Self> {
+        use std::fs;
+
+        let content = fs::read_to_string(path)
+            .with_context(|| format!("Could not read manifest file at {}", path.display()))?;
+
+        let manifest: Self = toml::from_str(&content)
+            .with_context(|| format!("Could not decode the manifest file at {}", path.display()))?;
```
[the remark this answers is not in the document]
KokaKiwi: It's already provided by `anyhow` which include the parent errors

crates/cli/src/components/manifest.rs
KokaKiwi: The field is already optional and I took the same behaviour as the console, which makes it empty when not provided

crates/cli/src/commands/components/mod.rs
```diff
@@ -0,0 +1,27 @@
+setup_commands! {
+    /// Init a new component in the current directory
```
alexcasalboni, line 2: "Init a new component manifest in the current directory"

crates/cli/src/commands/components/mod.rs
```diff
@@ -0,0 +1,27 @@
...
+    Init(init),
+    /// Create a new component in a new directory
+    New(new),
+
+    /// Build the component
```
alexcasalboni, line 7: "Compile the component in the current directory into Wasm"

crates/cli/src/commands/components/mod.rs
```diff
@@ -0,0 +1,27 @@
...
+    Build(build),
+
+    /// Pull a component
```
alexcasalboni, line 10: "Pull a component from the Edgee Component Registry"

crates/cli/src/commands/components/mod.rs
```diff
@@ -0,0 +1,27 @@
...
+    Pull(pull),
+    /// Push a component
```
alexcasalboni, line 12: "Push a component to the Edgee Component Registry"

crates/cli/src/commands/components/mod.rs
```diff
@@ -0,0 +1,27 @@
...
+    Push(push),
+    /// List currently pulled components
+    List(list),
+    /// Check if a component implements the required interfaces
```
alexcasalboni, line 16: "Check if the local .wasm file is valid"

crates/cli/src/commands/components/mod.rs
```diff
@@ -0,0 +1,27 @@
...
+    Check(check),
+    /// Test a component
```
alexcasalboni, line 18: "Run the component in the current folder with sample events."

crates/cli/src/commands/components/mod.rs
```diff
@@ -0,0 +1,27 @@
...
+    New(new),
+
+    /// Build the component
+    Build(build),
+
+    /// Pull a component
+    Pull(pull),
+    /// Push a component
+    Push(push),
+    /// List currently pulled components
```
alexcasalboni, line 14: "List components you've previously pulled"

crates/cli/src/commands/components/mod.rs
```diff
@@ -0,0 +1,27 @@
+setup_commands! {
+    /// Init a new component in the current directory
+    Init(init),
+    /// Create a new component in a new directory
```
alexcasalboni, line 4: "Create component in a new directory with sample code"

crates/cli/src/commands/mod.rs
```diff
@@ -4,6 +4,9 @@ mod macros;
 setup_commands! {
     #[command(flatten)]
     Auth(auth),
+    /// Components management commands
+    #[command(subcommand, alias = "component")]
```
alexcasalboni, line 8: I think it's ok to use`visible_alias` here 👌

crates/cli/src/commands/components/push.rs
```diff
@@ -0,0 +1,104 @@
+#[derive(Debug, clap::Parser)]
+pub struct Options {
+    /// Which organization to create the component into if not existing already.
```
alexcasalboni, line 3: "The organization name used to create or update your component."

crates/cli/src/commands/components/test.rs
```diff
@@ -0,0 +1,173 @@
...
+
+use edgee_components_runtime::data_collection::payload::{Event, EventType};
+
+#[derive(Debug, clap::Parser)]
+pub struct Options {
+    /// Comma-separated key=value pairs for settings
+    #[arg(long="settings", value_parser = parse_settings)]
+    pub settings: Option<HashMap<String, String>>,
+
+    /// Test to run
```
alexcasalboni, line 13: "The event type you want to test (valid values: page, track, or user)"

crates/cli/src/commands/components/push.rs
```diff
@@ -102,3 +116,18 @@ pub async fn run(opts: Options) -> anyhow::Result<()> {
...
+
+fn convert_manifest_config_fields(manifest: &Manifest) -> Vec<api_types::ConfigurationField> {
+    manifest
+        .package
+        .config_fields
+        .iter()
+        .map(|(name, field)| api_types::ConfigurationField {
+            name: name.clone(),
+            title: field.title.clone(),
+            type_: field.type_,
```
alexcasalboni, line 128: why `type_`? 🤔

KokaKiwi: Because `type` is a keyword in Rust :)

crates/cli/src/commands/components/test.rs
CLEMENTINATOR: lets do it at a later time

alexcasalboni (approved):

crates/cli/src/commands/components/check.rs
```diff
@@ -0,0 +1,88 @@
...
+    Ok(())
+}
+
+pub async fn run(_opts: Options) -> anyhow::Result<()> {
+    use anyhow::Context;
+
+    use crate::components::manifest::{self, Manifest};
+
+    let Some(manifest_path) = manifest::find_manifest_path() else {
+        anyhow::bail!("Edgee Manifest not found. Please run `edgee component create` and start from a template or `edgee component init` to create a new empty manifest in this folder.");
```
NicolasGirardot, line 73: isn't in edgee component new ?

NicolasGirardot (approved):

SachaMorard (approved):

PR #198, opened 2025-02-03
add Check command to check if a component is compatible with edgee

CLEMENTINATOR:
### Checklist

@KokaKiwi it is set to merge into your branch


CLEMENTINATOR:
tested with rs component

```
╭─clement@brittle-hollow ~/work/components/example-rs-component ‹initial_import●›
╰─$ ../../edgee/target/debug/edgee components check
╭─clement@brittle-hollow ~/work/components/example-rs-component ‹initial_import●›
╰─$ ../../edgee/target/debug/edgee components check -c target/wasm32-wasip2/release/example_rs_component.wasm
Error: no exported instance named `edgee:protocols/consent-mapping`
╭─clement@brittle-hollow ~/work/components/example-rs-component ‹initial_import●›
╰─$ ../../edgee/target/debug/edgee components check -d target/wasm32-wasip2/release/example_rs_component.wasm                                                                                                           
```

alexcasalboni:
Imho, the `Error: no exported instance named edgee:protocols/consent-mapping` error is very confusing and we should try to wrap the runtime errors into understandable messages like "Component not found", "Component not valid", "Invalid component type", "Component is valid", etc.

Also, would it be possible to infer the component type (data collection, consent mapping, etc.) without specifying it with `-c` or `-d`? That would make it much easier to use.

CLEMENTINATOR:
> Imho, the `Error: no exported instance named edgee:protocols/consent-mapping` error is very confusing and we should try to wrap the runtime errors into understandable messages like "Component not found", "Component not valid", "Invalid component type", "Component is valid", etc.
> 
> Also, would it be possible to infer the component type (data collection, consent mapping, etc.) without specifying it with `-c` or `-d`? That would make it much easier to use.

I'll wrap the errors !

Regarding the second question, I am not sure that we can detect the component type (directly from a wasm file i mean).

alexcasalboni:
> Regarding the second question, I am not sure that we can detect the component type (directly from a wasm file i mean).

We should be able to inspect the interface and check what it's export. Something like this:
```
wasm-tools component wit component.wasm
```

At some point the output should contain `package edgee:protocols {
    ...
    page: func(e: event, cred: dict) -> result<edgee-request, string>;
    track: func(e: event, cred: dict) -> result<edgee-request, string>;
    user: func(e: event, cred: dict) -> result<edgee-request, string>;
}`

CLEMENTINATOR:
> > Regarding the second question, I am not sure that we can detect the component type (directly from a wasm file i mean).
> 
> We should be able to inspect the interface and check what it's export. Something like this:
> 
> ```
> wasm-tools component wit component.wasm
> ```
> 
> At some point the output should contain
> 
> ```
> package edgee:protocols {
>     interface data-collection {
>         ...
>         page: func(e: event, cred: dict) -> result<edgee-request, string>;
>         track: func(e: event, cred: dict) -> result<edgee-request, string>;
>         user: func(e: event, cred: dict) -> result<edgee-request, string>;
>     }
> }
> ```

ill check what i can do then

CLEMENTINATOR:
> > Regarding the second question, I am not sure that we can detect the component type (directly from a wasm file i mean).
> 
> We should be able to inspect the interface and check what it's export. Something like this:
> 
> ```
> wasm-tools component wit component.wasm
> ```
> 
> At some point the output should contain
> 
> ```
> package edgee:protocols {
>     interface data-collection {
>         ...
>         page: func(e: event, cred: dict) -> result<edgee-request, string>;
>         track: func(e: event, cred: dict) -> result<edgee-request, string>;
>         user: func(e: event, cred: dict) -> result<edgee-request, string>;
>     }
> }
> ```

For me, this tool checks that the component implements a given world (and not the world that is in the component itself)

Eg if an user creates a cmp component, and defines it as a data-collection component in the registry, the tool should check the component against the data-collection component type and not the cmp component type (the one that is detected by wasm tools component wit)

crates/cli/src/commands/components/check.rs
```diff
@@ -0,0 +1,86 @@
...
+        manifest::find_manifest_path().ok_or_else(|| anyhow::anyhow!("Manifest not found"))?;
+
+    let manifest = Manifest::load(&manifest_path)?;
+    let component_path = manifest
+        .package
+        .build
+        .output_path
+        .into_os_string()
+        .into_string()
+        .map_err(|_| anyhow::anyhow!("Invalid path"))?;
```
alexcasalboni, line 80: what path? I suggest adding some more debugging info when printing errors (with a hint for instructions for how to fix the error).

Something like:

```
No path found in your Manifest file. Please add this to your manifest: "..."
```

crates/cli/src/commands/components/check.rs
```diff
@@ -0,0 +1,86 @@
...
+
+    let context = ComponentsContext::new(&config)
+        .map_err(|e| anyhow::anyhow!("Invalid component {}: {}", component_path, e))?;
+
+    let mut store = context.empty_store();
+
+    match component_type {
+        ComponentType::DataCollection => {
+            let _ = context
+                .get_data_collection_instance(component_path, &mut store)
```
alexcasalboni, line 53: what error message is printed if this fails?

crates/cli/src/commands/components/check.rs
```diff
@@ -0,0 +1,86 @@
...
+
+    println!("Component {} is valid", component_path);
+    Ok(())
+}
+
+pub async fn run(_opts: Options) -> anyhow::Result<()> {
+    use crate::components::manifest::{self, Manifest};
+
+    let manifest_path =
+        manifest::find_manifest_path().ok_or_else(|| anyhow::anyhow!("Manifest not found"))?;
```
alexcasalboni, line 71: Same here.

I'd expect some useful hints to fix the error.

```
Edgee Manifest not found. Please run `edgee component create` and start from a template or `edgee component init` to create a new empty manifest in this folder.
```

crates/cli/src/commands/components/check.rs
CLEMENTINATOR: I'm actually unsure on how to make it fail lol

alexcasalboni: mmm because it would fail in line 45-46 anyway?

CLEMENTINATOR: yeah tehre are many errors that can occur within lines 45-46

alexcasalboni (approved):

alexcasalboni:
I think this command isn't going to be used by OSS developers too often and we probably want to automatically run it before `edgee component push` to make sure the .wasm file is valid before uploading it.

CLEMENTINATOR:
> I think this command isn't going to be used by OSS developers too often and we probably want to automatically run it before `edgee component push` to make sure the .wasm file is valid before uploading it.

Exactly, its to precheck on the client side before we check it ourself on server side !

PR #199, opened 2025-02-04
feat: new component configuration

SachaMorard:
### Checklist

### Description of Changes

This pull request introduces changes to DataCollectionComponents configuration fields. 
- `name` becomes `id`
- `component` becomes `file`
- `credentials` and `config` are merged into `settings`
- `forward_client_headers` is removed (it is now a part of the `edgee_request` object return by components)

These changes affect edgee.toml components part:

```toml
[[components.data_collection]]
id = "amplitude" # could be a slug (edgee/amplitude) or an alias (amplitude)
file = "../components/amplitude-component/amplitude.wasm"
settings.amplitude_api_key = "xxxx"
settings.edgee_anonymization = true
settings.edgee_default_consent = "pending"
settings.edgee_track_event_enabled = false
settings.edgee_user_event_enabled = true
settings.edgee_page_event_enabled = true
```

##### Other changes:
- New wit protocol (with `forward_client_headers`)
- Openapi.json fixes
- Readme changes following the new component configuration

### Related Issues

List related issues here

### DIFF
--- a/crates/components-runtime/src/data_collection/mod.rs
+++ b/crates/components-runtime/src/data_collection/mod.rs
@@ -143,7 +143,10 @@ pub async fn send_events(
             }
 
             // get the instance of the component
-            let instance = match component_ctx.get_data_collection_instance(&cfg.id, &mut store).await {
+            let instance = match component_ctx
+                .get_data_collection_instance(&cfg.id, &mut store)
+                .await
+            {
                 Ok(instance) => instance,
                 Err(err) => {
                     error!("Failed to get data collection instance. Error: {}", err);
--- a/crates/components-runtime/src/data_collection/mod.rs
+++ b/crates/components-runtime/src/data_collection/mod.rs
@@ -201,7 +201,7 @@ pub async fn send_events(
                 headers.insert(HeaderName::from_str(key)?, HeaderValue::from_str(value)?);
             }
 
-            if cfg.forward_client_headers.unwrap_or(true) {
+            if request.forward_client_headers {
                 insert_expected_headers(&mut headers, &event, &component_event)?;
             }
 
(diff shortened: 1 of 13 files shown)


crates/components-runtime/src/data_collection/payload.rs
```diff
@@ -98,10 +100,51 @@ impl Event {
             .get("all")
             .unwrap_or(&true);
 
-        // check if the components is enabled
-        if self.components.as_ref().unwrap().contains_key(name) {
-            return self.components.as_ref().unwrap().get(name).unwrap();
+        // check if the components is enabled by id
+        if self
```
CLEMENTINATOR, line 104: nitpick:
```
if let Some(components) = self.components.as_ref() {
    if let Some(component) = components.get(config.id.as_str()) {
        return component;
    }

    if let Some(component) = components.get(config.project_component_id.as_str()) {
        return component;
    }

    if let Some(component) = components.get(config.slug.as_str()) {
        return component;
    }
}
```

CLEMENTINATOR (approved):
lgtm otherwise

crates/components-runtime/src/data_collection/mod.rs
```diff
@@ -195,7 +201,7 @@ pub async fn send_events(
                 headers.insert(HeaderName::from_str(key)?, HeaderValue::from_str(value)?);
             }
 
-            if cfg.forward_client_headers.unwrap_or(true) {
+            if request.forward_client_headers {
```
alexcasalboni, line 204: Generally, I would expect this to happen either always or never for a given component (so we initially implemented it as a static setting). But it also makes sense to set it dynamically for each EdgeeRequest, for more flexibility 👍

alexcasalboni (approved):
LGTM 💯

crates/components-runtime/src/data_collection/payload.rs
SachaMorard: @CLEMENTINATOR I found a better way

CLEMENTINATOR: :+1:

PR #204, opened 2025-02-07
feat: add new and init command

NicolasGirardot:
### Checklist

### Description of Changes

Describe your changes here

### Related Issues

List related issues here

### DIFF
--- a/crates/cli/Cargo.toml
+++ b/crates/cli/Cargo.toml
@@ -23,8 +23,8 @@ tokio = { workspace = true, features = ["macros", "rt-multi-thread"] }
 toml.workspace = true
 tracing-subscriber = { workspace = true, features = ["env-filter", "json"] }
 url = { workspace = true, features = ["serde"] }
-reqwest = { version = "0.11", features = ["blocking"] }
-zip = "2.2.2"
+reqwest = { workspace = true, features = ["blocking"] }
+zip.workspace = true
 
 edgee-api-client.workspace = true
 edgee-server.workspace = true
--- a/crates/cli/Cargo.toml
+++ b/crates/cli/Cargo.toml
@@ -23,12 +23,12 @@ tokio = { workspace = true, features = ["macros", "rt-multi-thread"] }
 toml.workspace = true
 tracing-subscriber = { workspace = true, features = ["env-filter", "json"] }
 url = { workspace = true, features = ["serde"] }
+reqwest = { version = "0.11", features = ["blocking"] }
+zip = "2.2.2"
 
 edgee-api-client.workspace = true
 edgee-server.workspace = true
 edgee-components-runtime.workspace = true
 
 [features]
-bundled = [
-    "openssl/vendored",
-]
+bundled = ["openssl/vendored"]
--- a/crates/cli/src/commands/components/init.rs
+++ b/crates/cli/src/commands/components/init.rs
@@ -1,5 +1,5 @@
 use crate::components::{
-    boilerplates::LANGUAGE_OPTIONS,
+    boilerplates::{CATEGORY_OPTIONS, LANGUAGE_OPTIONS, SUBCATEGORY_OPTIONS},
     manifest::{self, Build, Manifest, Package},
 };
 
@@ -21,6 +21,21 @@ pub async fn run(_opts: Options) -> anyhow::Result<()> {
         LANGUAGE_OPTIONS.to_vec(),
     )
     .prompt()?;
+    let component_category = if CATEGORY_OPTIONS.len() == 1 {
+        CATEGORY_OPTIONS[0].clone() // Accès direct car on sait qu'il y a un seul élément
+    } else {
+        Select::new(
+            "Select the category of the component:",
+            CATEGORY_OPTIONS.to_vec(), // Pas besoin de `.to_vec()`, on passe une slice
+        )
+        .prompt()?
+    };
+
+    let component_subcategory = Select::new(
+        "Select the language of the component:",
+        SUBCATEGORY_OPTIONS.to_vec(),
+    )
+    .prompt()?;
 
     println!(
         "Initiating component {} in {}",
@@ -33,6 +48,11 @@ pub async fn run(_opts: Options) -> anyhow::Result<()> {
             name: component_name,
             version: "0.1.0".to_string(),
             wit_world_version: "0.4.0".to_string(),
+            category: *component_category.value,
+            subcategory: *component_subcategory.value,
+            description: None,
+            documentation: None,
+            repository: None,
             build: Build {
                 command: component_language.default_build_command.to_string(),
                 output_path: std::path::PathBuf::from(""),
--- a/crates/cli/src/commands/components/init.rs
+++ b/crates/cli/src/commands/components/init.rs
@@ -57,6 +57,7 @@ pub async fn run(_opts: Options) -> anyhow::Result<()> {
             description: None,
             documentation: None,
             repository: None,
+            config_fields: Default::default(),
             build: Build {
                 command: component_language.default_build_command.to_string(),
                 output_path: std::path::PathBuf::from(""),
--- a/crates/cli/src/commands/components/init.rs
+++ b/crates/cli/src/commands/components/init.rs
@@ -8,7 +8,7 @@ pub struct Options {}
 
 pub async fn run(_opts: Options) -> anyhow::Result<()> {
     use inquire::{Select, Text};
-    if let Some(_) = manifest::find_manifest_path() {
+    if manifest::find_manifest_path().is_some() {
         anyhow::bail!("Manifest already exists");
     }
 
--- a/crates/cli/src/commands/components/init.rs
+++ b/crates/cli/src/commands/components/init.rs
@@ -1,5 +1,5 @@
 use crate::components::{
-    boilerplates::{CATEGORY_OPTIONS, LANGUAGE_OPTIONS, SUBCATEGORY_OPTIONS},
+    boilerplate::{CATEGORY_OPTIONS, LANGUAGE_OPTIONS, SUBCATEGORY_OPTIONS},
     manifest::{self, Build, Manifest, Package},
 };
 
@@ -14,6 +14,10 @@ pub async fn run(_opts: Options) -> anyhow::Result<()> {
 
     let component_name = Text::new("Enter the name of the component:")
         .with_validator(inquire::required!("Component name cannot be empty"))
+        .with_validator(inquire::min_length!(
+            3,
+            "Component name must be at least 3 characters"
+        ))
         .prompt()?;
 
     let component_language = Select::new(
--- a/crates/cli/src/commands/components/init.rs
+++ b/crates/cli/src/commands/components/init.rs
@@ -32,9 +32,9 @@ pub async fn run(_opts: Options) -> anyhow::Result<()> {
         package: Package {
             name: component_name,
             version: "0.1.0".to_string(),
-            wit_world_version: "0.1.0".to_string(),
+            wit_world_version: "0.4.0".to_string(),
             build: Build {
-                command: "".to_string(),
+                command: component_language.default_build_command.to_string(),
                 output_path: std::path::PathBuf::from(""),
             },
         },
--- a/crates/cli/src/commands/components/init.rs
+++ b/crates/cli/src/commands/components/init.rs
@@ -1,6 +1,45 @@
+use crate::components::{
+    boilerplates::LANGUAGE_OPTIONS,
+    manifest::{self, Build, Manifest, Package},
+};
+
 #[derive(Debug, clap::Parser)]
 pub struct Options {}
 
 pub async fn run(_opts: Options) -> anyhow::Result<()> {
+    use inquire::{Select, Text};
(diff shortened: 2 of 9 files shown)


crates/cli/src/commands/components/init.rs
```diff
@@ -1,6 +1,45 @@
...
 pub struct Options {}
 
 pub async fn run(_opts: Options) -> anyhow::Result<()> {
+    use inquire::{Select, Text};
+    if let Some(_) = manifest::find_manifest_path() {
+        anyhow::bail!("Manifest already exists");
+    }
+
+    let component_name = Text::new("Enter the name of the component:")
+        .with_validator(inquire::required!("Component name cannot be empty"))
```
alexcasalboni, line 16: any other requirement? can it be just 1 character?

crates/cli/src/commands/components/init.rs
```diff
@@ -1,6 +1,45 @@
...
+        .prompt()?;
+
+    let component_language = Select::new(
+        "Select the language of the component:",
```
alexcasalboni, line 20: a comment about the future: at some point we will have multiple types of components, so do we expect to have different templates based on language+type or can we just keep 1 boilerplace repo per language that includes all types?

if we go with multiple repos based on language+types, here we will need to select the language/type combo (or maybe an additional step to select the type, and then the language).

If we go for just language selection, this is already OK 👌

crates/cli/src/commands/components/init.rs
```diff
@@ -1,6 +1,45 @@
...
+        "Initiating component {} in {}",
+        component_name, component_language.name
+    );
+
+    Manifest {
+        manifest_version: manifest::MANIFEST_VERSION,
+        package: Package {
+            name: component_name,
+            version: "0.1.0".to_string(),
+            wit_world_version: "0.1.0".to_string(),
```
alexcasalboni, line 35: should we try to use the latest here?

crates/cli/src/commands/components/init.rs
```diff
@@ -1,6 +1,45 @@
...
+            build: Build {
+                command: "".to_string(),
```
alexcasalboni, line 37: no default build command based on the selected language?

crates/cli/src/commands/components/new.rs
```diff
@@ -1,6 +1,63 @@
...
+        LANGUAGE_OPTIONS.to_vec(),
+    )
+    .prompt()?;
+
+    let component_path = Path::new(&component_name);
+    if component_path.exists() {
+        anyhow::bail!("A component with this name already exists in this directory");
+    }
+
+    let url = format!(
```
alexcasalboni, line 30: I'd add something like:

```
println!(
    "Downloading sample code for {}", component_language.name
);
```

crates/cli/src/commands/components/new.rs
```diff
@@ -1,6 +1,63 @@
...
+        "{}{}",
+        component_language.repo_url, "/archive/refs/heads/main.zip"
+    );
+    let response = Client::new().get(url).send().await?.bytes().await?;
+    let reader = Cursor::new(response);
+    let mut archive = ZipArchive::new(reader)?;
+
+    for i in 0..archive.len() {
```
alexcasalboni, line 38: And then:

```
println!(
    "Creating new folder for {}", component_language.name
);
```

or something like that - just to show some progress

crates/cli/src/components/boilerplates.rs
```diff
@@ -0,0 +1,38 @@
...
+    pub repo_url: &'static str,
+}
+
+impl std::fmt::Display for LanguageConfig {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self.name)
+    }
+}
+
+pub const LANGUAGE_OPTIONS: &'static [LanguageConfig] = &[
```
alexcasalboni: should these be sorted alphabetically? or some other sorting logic that makes sense for us?

crates/cli/Cargo.toml
```diff
@@ -22,12 +22,12 @@ tracing.workspace = true
 tokio = { workspace = true, features = ["macros", "rt-multi-thread"] }
 toml.workspace = true
 tracing-subscriber = { workspace = true, features = ["env-filter", "json"] }
+reqwest = { version = "0.11", features = ["blocking"] }
+zip = "2.2.2"
```
KokaKiwi, line 26: Please define dependencies versions in the root `Cargo.toml` and use the `workspace` field on subcrates
This way it's way easier to track dependencies :)

KokaKiwi (changes requested):

crates/cli/src/commands/components/init.rs
CLEMENTINATOR: has to be 3 min

crates/cli/src/commands/components/init.rs
```diff
@@ -1,6 +1,45 @@
...
+        .with_validator(inquire::required!("Component name cannot be empty"))
+        .prompt()?;
+
+    let component_language = Select::new(
+        "Select the language of the component:",
+        LANGUAGE_OPTIONS.to_vec(),
+    )
+    .prompt()?;
+
+    println!(
```
CLEMENTINATOR, line 25: @KokaKiwi is there any crates for printing that you might recommand ?

crates/cli/src/commands/components/init.rs
```diff
@@ -1,6 +1,45 @@
...
+        "Initiating component {} in {}",
+        component_name, component_language.name
+    );
+
+    Manifest {
+        manifest_version: manifest::MANIFEST_VERSION,
+        package: Package {
```
CLEMENTINATOR, line 32: as clippy said you're missing a lot of fields :D

crates/cli/src/components/boilerplates.rs
```diff
@@ -0,0 +1,45 @@
...
+    },
+    LanguageConfig {
+        name: "Go",
+        repo_url: "https://github.com/edgee-cloud/example-go-component",
+        default_build_command: "go build -o main .",
+    },
+    LanguageConfig {
+        name: "Python",
+        repo_url: "https://github.com/edgee-cloud/example-py-component",
+        default_build_command: "python main.py",
```
CLEMENTINATOR, line 28: is that the command for building the python component ? :thinking:

crates/cli/src/components/boilerplates.rs
```diff
@@ -0,0 +1,45 @@
...
+pub static LANGUAGE_OPTIONS: &[LanguageConfig] = &[
+    LanguageConfig {
+        name: "Rust",
+        repo_url: "https://github.com/edgee-cloud/example-rust-component",
+        default_build_command: "cargo build --release",
+    },
+    LanguageConfig {
+        name: "Go",
+        repo_url: "https://github.com/edgee-cloud/example-go-component",
+        default_build_command: "go build -o main .",
```
CLEMENTINATOR, line 23: ditto for go

crates/cli/src/components/boilerplates.rs
```diff
@@ -0,0 +1,45 @@
...
+    },
+    LanguageConfig {
+        name: "Python",
+        repo_url: "https://github.com/edgee-cloud/example-py-component",
+        default_build_command: "python main.py",
+    },
+    LanguageConfig {
+        name: "JavaScript",
+        repo_url: "https://github.com/edgee-cloud/example-js-component",
+        default_build_command: "node main.js",
```
CLEMENTINATOR, line 33: same here :thinking:

crates/cli/src/components/boilerplates.rs
```diff
@@ -0,0 +1,45 @@
...
+    pub default_build_command: &'static str,
+}
+
+impl std::fmt::Display for LanguageConfig {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self.name)
+    }
+}
+
+pub static LANGUAGE_OPTIONS: &[LanguageConfig] = &[
```
CLEMENTINATOR, line 14: @alexcasalboni i think we're missing instructions on how to install specific dependencies for languages, wdyt ?

crates/cli/src/components/mod.rs
```diff
@@ -1 +1,2 @@
+pub mod boilerplates;
```
CLEMENTINATOR, line 1: mod should be named boilerplate without plural imo

crates/cli/src/commands/components/init.rs
CLEMENTINATOR: we'd still need to have a subcategory selection for the component

crates/cli/src/components/boilerplates.rs
alexcasalboni: I would expect those to be well-documented in the sample README 🤔 but we could add a simple reminder in the output of the init/new command, something like "Don't forget to install local dependencies before building".

crates/cli/src/components/boilerplates.rs
alexcasalboni: is what we have in the go sample repo ok?

```go
tinygo build -target=wasip2 -o dc_component.wasm --wit-package wit/ --wit-world data-collection ./
```

I also wonder if we should support something like dynamic parameters in those scripts, like `tinygo build -target=wasip2 -o ${output_path} --wit-package wit/ --wit-world data-collection ./` so that we can actually use the `output_path` without defining it multiple times

crates/cli/src/components/boilerplates.rs
alexcasalboni: it's fair to assume there's an `npm run build` script?

since the JS tooling depends on a specific `build.mjs` file, can we do anything about that?


@CLEMENTINATOR I see that ComponentizeJS can also be run as a CLI command, like this:
```bash
jco componentize source.js --wit wit -o component.wasm
```

so maybe it's better to change the default/sample implementation to use that instead of depending on a build.mjs file that wouldn't exist if you just run `edgee component init`?

crates/cli/src/components/boilerplates.rs
NicolasGirardot: So I thought about this, this is only for the init method since the new components include the manifest in the repo. 

In that situation I'm not aware on how the user is gonna code his component, neither if it's gonna be a data collection or simply something else. Not sure what we are suppose to define as "The defautl command" that why I put the basic compilation of the language

NicolasGirardot: in my opinion (but I could be very wrong) the user should define their command in the case of init, we're not aware of what you're about to do

crates/cli/src/commands/components/init.rs
NicolasGirardot: Updated it :D

crates/cli/src/commands/components/init.rs
NicolasGirardot: I've added it following Jill changes to include it in manifest

crates/cli/src/components/boilerplates.rs
alexcasalboni: they'll definitely need to customize it if they are using something out of the ordinary (based on the official Wasm/WASI specs), but we can definitely provide a good-enough default to start from

crates/cli/src/components/boilerplates.rs
```diff
@@ -0,0 +1,45 @@
...
+    },
+    LanguageConfig {
+        name: "CSharp",
+        repo_url: "https://github.com/edgee-cloud/example-csharp-component",
+        default_build_command: "dotnet build",
+    },
+    LanguageConfig {
+        name: "C",
+        repo_url: "https://github.com/edgee-cloud/example-c-component",
+        default_build_command: "gcc main.c -o main",
```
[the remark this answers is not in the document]
alexcasalboni: would this work?

```bash
$(CC) dc_component.c internal/data_collection.c internal/data_collection_component_type.o -o dc_component_temp.wasm -mexec-model=reactor -Os | wasm-tools component new -o dc_component.wasm
```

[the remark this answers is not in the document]
CLEMENTINATOR: wasm-tools can take stdin but cc does not output the file :/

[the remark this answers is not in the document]
CLEMENTINATOR: ```-o /dev/stdout``` can try with that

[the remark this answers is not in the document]
CLEMENTINATOR: i updated the c repo with the correct command

crates/cli/src/commands/components/new.rs
```diff
@@ -1,6 +1,63 @@
+use reqwest::Client;
+use std::fs::{create_dir_all, File};
+use std::io::{Cursor, Read, Write};
+use std::path::Path;
+use zip::read::ZipArchive;
+
+use crate::components::boilerplates::LANGUAGE_OPTIONS;
+
 #[derive(Debug, clap::Parser)]
 pub struct Options {}
```
alexcasalboni, line 10: ideally we would also add optional parameters here so that name and language can be provided as input parameters 👌  (could be in the next PR)

PR #209, opened 2025-02-13
feat: add all field as example value when initing

NicolasGirardot:
Generated Manifest
image

### Checklist

### Description of Changes

Describe your changes here

### Related Issues

List related issues here


NicolasGirardot:
Blocked by #207 since it will change the name of the config_field to setting

crates/cli/src/commands/components/init.rs
```diff
@@ -54,10 +56,22 @@ pub async fn run(_opts: Options) -> anyhow::Result<()> {
...
-            config_fields: Default::default(),
+            description: Some("Description of the component".to_string()),
+            documentation: Some(Url::parse("https://www.edgee.cloud/docs/introduction")?),
+            repository: Some(Url::parse("https://www.github.com/edgee-cloud/edgee")?),
+            config_fields: {
+                let mut fields = std::collections::HashMap::new();
+                fields.insert(
+                    "example".to_string(),
+                    ConfigField {
+                        description: Some("Example config field".to_string()),
```
alexcasalboni, line 67: it will be "Example setting" once the rename is complete 👌

crates/cli/src/commands/components/init.rs
```diff
@@ -54,10 +56,22 @@ pub async fn run(_opts: Options) -> anyhow::Result<()> {
             wit_world_version: "0.4.0".to_string(),
             category: *component_category.value,
             subcategory: *component_subcategory.value,
-            description: None,
-            documentation: None,
-            repository: None,
-            config_fields: Default::default(),
+            description: Some("Description of the component".to_string()),
```
alexcasalboni, line 59: it would be great to provide a multi-line string by default :) like this:

```toml
description = '''
Here is a 
string
'''
```

NicolasGirardot: image
Updated !

alexcasalboni: is there a way to force [component.build] to appear before the settings? or is that done automatically during serialization?

NicolasGirardot: I changed the order in the object class definition

alexcasalboni (approved):

PR #210, opened 2025-02-13
Prompt private or public on creation

CLEMENTINATOR:
### Checklist

### Description of Changes

push: prompt for public or private on creation

### Related Issues

List related issues here

### DIFF
--- a/crates/cli/src/commands/components/push.rs
+++ b/crates/cli/src/commands/components/push.rs
@@ -10,7 +10,7 @@ pub struct Options {
 }
 
 pub async fn run(opts: Options) -> anyhow::Result<()> {
-    use inquire::{Confirm, Editor};
+    use inquire::{Confirm, Editor, Select};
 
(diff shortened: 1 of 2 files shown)


crates/cli/src/commands/components/push.rs
```diff
@@ -64,6 +64,12 @@ pub async fn run(opts: Options) -> anyhow::Result<()> {
                 return Ok(());
             }
 
+            let public = Select::new(
```
NicolasGirardot, line 67: So we agree that it will get prompted every push ?

alexcasalboni: this is only when the component doesn't exist yet

alexcasalboni (approved):
LGTM 👍

crates/cli/src/commands/components/push.rs
NicolasGirardot: Ah yeah didn't see the code above

NicolasGirardot (approved):

PR #212, opened 2025-02-13
components test: check that all required settings are provided

CLEMENTINATOR:
### Checklist

### Description of Changes

components test: check that all required settings are provided


crates/cli/src/commands/components/test.rs
```diff
@@ -72,9 +87,15 @@ async fn test_data_collection_component(component_path: &str, opts: Options) ->
             settings_map.insert(key, value);
         }
     }
-    // TODO generate settings from the missing settings in the component manifest
-    let settings = settings_map.clone().into_iter().collect();
 
+    // check that all required settings are provided
+    for (name, setting) in manifest.package.settings {
+        if setting.required && settings_map.get(&name) == None {
```
alexcasalboni, line 93: I would also return an error in case they provide a setting that isn't defined in the manifest

PR #216, opened 2025-02-13
feat(cli/components): improve push subcommand

KokaKiwi:
### Checklist

### Description of Changes

- Build component on push when not built
- Update component metadata on push
- Print component URL on push

### Related Issues

- Complete EDGEE-444, EDGEE-445 and EDGEE-455

### DIFF
--- a/crates/cli/src/commands/components/push.rs
+++ b/crates/cli/src/commands/components/push.rs
@@ -106,7 +106,33 @@ pub async fn run(opts: Options) -> anyhow::Result<()> {
                 component_slug
             );
         }
-        Ok(_) | Err(_) => {}
+        Ok(_) | Err(_) => {
+            client
+                .update_component_by_slug()
+                .org_slug(&organization.slug)
+                .component_slug(&manifest.component.name)
+                .body(
+                    api_types::ComponentUpdateParams::builder()
+                        .description(manifest.component.description.clone())
+                        .documentation_link(
+                            manifest
+                                .component
+                                .documentation
+                                .as_ref()
+                                .map(|url| url.to_string()),
+                        )
+                        .repo_link(
+                            manifest
+                                .component
+                                .repository
+                                .as_ref()
+                                .map(|url| url.to_string()),
+                        ),
+                )
+                .send()
+                .await
+                .api_context("Could not update component infos")?;
+        }
     }
 
     let changelog =
--- a/crates/cli/src/commands/components/push.rs
+++ b/crates/cli/src/commands/components/push.rs
@@ -45,7 +45,7 @@ pub async fn run(opts: Options) -> anyhow::Result<()> {
 
     let component_slug = slug::slugify(&manifest.component.name);
     let component_url = format!(
-        "https://www.edgee.cloud/~/registry/{}/{}",
+        "https://www.edgee.cloud/~/{0}/component-registry/{0}/{1}",
         organization.slug, component_slug,
     );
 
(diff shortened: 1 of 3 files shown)


crates/cli/src/commands/components/push.rs
```diff
@@ -41,7 +42,13 @@ pub async fn run(opts: Options) -> anyhow::Result<()> {
             .api_context("Could not get user organization")?
             .into_inner(),
     };
+
+    let component_url = format!(
+        "https://www.edgee.cloud/~/registry/{}/{}",
+        organization.slug, manifest.component.name,
```
CLEMENTINATOR, line 48: use slug instead of name

crates/cli/src/commands/components/push.rs
```diff
@@ -100,12 +107,43 @@ pub async fn run(opts: Options) -> anyhow::Result<()> {
...
                 organization.slug,
                 component_slug
             );
         }
-        Ok(_) | Err(_) => {}
+        Ok(_) | Err(_) => {
+            client
+                .update_component_by_slug()
+                .org_slug(&organization.slug)
+                .component_slug(&manifest.component.name)
```
CLEMENTINATOR, line 119: name is not slug

crates/cli/src/commands/components/push.rs
KokaKiwi: Woops, indeed

CLEMENTINATOR (approved):
lgtm !

crates/cli/src/commands/components/push.rs
```diff
@@ -41,7 +42,13 @@ pub async fn run(opts: Options) -> anyhow::Result<()> {
             .api_context("Could not get user organization")?
             .into_inner(),
     };
+
     let component_slug = slug::slugify(&manifest.component.name);
+    let component_url = format!(
+        "https://www.edgee.cloud/~/registry/{}/{}",
```
alexcasalboni, line 48: until we update the UI, this needs to be:

```rust
let component_url = format!(
        "https://www.edgee.cloud/~/{}/component-registry/{}/{}",
        organization.slug, organization.slug, component_slug,
    );
```

crates/cli/src/commands/components/push.rs
```diff
@@ -100,12 +107,43 @@ pub async fn run(opts: Options) -> anyhow::Result<()> {
                 .await
                 .api_context("Could not create component")?;
             tracing::info!(
-                "Component `{}/{}` created successfully!",
+                "Component `{}/{}` has been created successfully, you check it out here: {component_url}",
```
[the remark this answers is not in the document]
alexcasalboni: Also, I would print the URL at the very end (after all the other logs below)

crates/cli/src/commands/components/push.rs
```diff
@@ -100,12 +107,43 @@ pub async fn run(opts: Options) -> anyhow::Result<()> {
...
+                                .repository
+                                .as_ref()
+                                .map(|url| url.to_string()),
+                        ),
+                )
+                .send()
+                .await
+                .api_context("Could not update component infos")?;
+            tracing::info!(
+                "Component `{}/{}` has been updated and is accessible here: {component_url}",
```
alexcasalboni, line 142: Same! let's print out the URL only at the very end of the command :) here we haven't even asked push confirmation 😆  maybe we should move the update logic AFTER asking for confirmation?

alexcasalboni (approved):
Let's go 🚀

PR #229, opened 2025-02-18
component test: add a --curl switch to get a curl command to test the output

CLEMENTINATOR:
### Checklist

### Description of Changes

component test: add a --curl switch to get a curl command to test the output

### DIFF
--- a/crates/cli/src/commands/components/test.rs
+++ b/crates/cli/src/commands/components/test.rs
@@ -4,6 +4,8 @@ use edgee_components_runtime::config::{ComponentsConfiguration, DataCollectionCo
 use edgee_components_runtime::context::ComponentsContext;
 use std::collections::HashMap;
 
+use edgee_components_runtime::data_collection::exports::edgee::protocols::data_collection::EdgeeRequest;
+use edgee_components_runtime::data_collection::exports::edgee::protocols::data_collection::HttpMethod;
 use edgee_components_runtime::data_collection::payload::{Event, EventType};
 
 #[derive(Debug, clap::Parser)]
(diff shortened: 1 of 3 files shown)


crates/cli/src/commands/components/test.rs
```diff
@@ -19,6 +21,37 @@ pub struct Options {
...
+            HttpMethod::Post => "POST".to_string(),
+            HttpMethod::Put => "PUT".to_string(),
+            HttpMethod::Delete => "DELETE".to_string(),
+        }
+    }
+}
+
+impl IntoCurl for EdgeeRequest {
+    fn to_curl(self) -> String {
+        let mut curl = format!("curl -X {} {}", self.method.to_curl(), self.url);
```
alexcasalboni, line 46: this is probably enough?

```Rust
self.method.to_uppercase()
```

alexcasalboni: right 😆  funny because it is a string in the JS/TS generated code, I got confused

crates/cli/src/commands/components/test.rs
```diff
@@ -165,25 +205,126 @@ async fn test_data_collection_component(opts: Options) -> anyhow::Result<()> {
...
+        .timeout(Duration::from_secs(5))
+        .build()?;
+
+    let mut headers = HeaderMap::new();
+    for (key, value) in request.headers.iter() {
+        headers.insert(HeaderName::from_str(key)?, HeaderValue::from_str(value)?);
+    }
+
+    let res = match request.method {
+        HttpMethod::Get => client.get(request.url).headers(headers).send().await,
```
alexcasalboni, line 267: could avoid some duplicated code here with the official `http::request::Builder`- should look like this:
```rust
Builder::new()
    .method(request.method.to_curl())
    .headers(headers)
    .body(request.body)
    .unwrap()
    .send().await
```

crates/cli/src/commands/components/test.rs
```diff
@@ -165,25 +205,126 @@ async fn test_data_collection_component(opts: Options) -> anyhow::Result<()> {
...
+                .post(request.url)
+                .headers(headers)
+                .body(request.body)
+                .send()
+                .await
+        }
+        HttpMethod::Delete => client.delete(request.url).headers(headers).send().await,
+    };
+
+    tracing::info!("Output from HTTP request:");
```
alexcasalboni, line 287: tracing::info!("HTTP response:");

crates/cli/src/commands/components/test.rs
```diff
@@ -165,25 +205,126 @@ async fn test_data_collection_component(opts: Options) -> anyhow::Result<()> {
...
+    match res {
+        Ok(res) => {
+            println!("\n{}: {}", "Response".green(), res.status());
```
alexcasalboni, line 290:
```rust
println!("\n{}: {}", "Status".green(), res.status());
```

alexcasalboni (approved):
Looks great 👍  only some minor comments

crates/cli/src/commands/components/test.rs
CLEMENTINATOR: for now ive done it as we do in the component runtime !

PR #231, opened 2025-02-18
feat(cli-components-new): add possibility to use argument to shortcut command

NicolasGirardot:
https://github.com/user-attachments/assets/4b4baf07-7891-4f48-ab38-28d0425a8ae7

### Checklist

### Description of Changes

Describe your changes here

### Related Issues

List related issues here

### DIFF
--- a/crates/cli/src/commands/components/new.rs
+++ b/crates/cli/src/commands/components/new.rs
@@ -4,7 +4,7 @@ use std::io::{Cursor, Read, Write};
 use std::path::Path;
 use zip::read::ZipArchive;
 
-use crate::components::boilerplate::LANGUAGE_OPTIONS;
+use crate::components::boilerplate::{LanguageConfig, LANGUAGE_OPTIONS};
 
 #[derive(Debug, clap::Parser)]
 pub struct Options {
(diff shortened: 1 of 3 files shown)


crates/cli/src/components/boilerplate.rs
```diff
@@ -40,36 +41,43 @@ pub static LANGUAGE_OPTIONS: &[LanguageConfig] = &[
...
         name: "Python",
         repo_url: "https://github.com/edgee-cloud/example-py-component",
         default_build_command: "python main.py",
+        alias: &["py", "python"],
     },
     LanguageConfig {
         name: "Rust",
         repo_url: "https://github.com/edgee-cloud/example-rust-component",
         default_build_command: "cargo build --release",
+        alias: &["rs", "rust"],
```
CLEMENTINATOR, line 74: @alexcasalboni should we put rust to the top ?

crates/cli/src/commands/components/new.rs
```diff
@@ -7,20 +7,53 @@ use zip::read::ZipArchive;
...
+                .iter()
+                .find(|l| l.alias.contains(&language.to_lowercase().as_str()))
+            {
+                lang
+            } else {
+                tracing::info!(
+                    "Language '{}' not available. Please select from the list:",
+                    language
+                );
+                &Select::new(
```
CLEMENTINATOR, line 44: I feel like there is a way to not duplicate code here

CLEMENTINATOR (approved):

crates/cli/src/components/boilerplate.rs
alexcasalboni: these are intentionally sorted alphabetically 👌  imho it makes it faster to find what you're looking for (rather than a custom order based on our preference/priority)

alexcasalboni (approved):

crates/cli/src/commands/components/new.rs
NicolasGirardot: image
Let me know what you think !

CLEMENTINATOR: prompt language is only called once, i think you could drop the let prompt language :thinking: but rest is perfect !

PR #234, opened 2025-02-19
feat(cli): add self-update subcommand

KokaKiwi:
### Checklist

### Description of Changes

Add a `self-update` command to automatically update the Edgee executable

### Related Issues

- EDGEE-238


crates/cli/src/commands/update.rs
```diff
@@ -0,0 +1,24 @@
+setup_command! {}
+
+pub async fn run(_opts: Options) -> anyhow::Result<()> {
+    // self_update uses synchronous reqwest client so we need to run it in a blocking task
+    tokio::task::spawn_blocking(move || {
+        use self_update::{backends::github::Update, Status};
+
+        let updater = Update::configure()
```
alexcasalboni, line 8: does this work only if you have used the install script? 

is there a way to detect alternative installation methods? like `homebrew`? or `WinGet` for Windows in the future? I would imagine this would break (or simply not work)?

KokaKiwi: I added a feature `no-self-update` to use when packaging for systems with their own package manager (linux distros, homebrew, etc.) which disable this subcommand

alexcasalboni (approved):
Cool 👍

PR #236, opened 2025-02-19
Component test forward headers

CLEMENTINATOR:
### Checklist

### Description of Changes

component test: forward client headers

### DIFF
--- a/crates/cli/src/commands/components/test.rs
+++ b/crates/cli/src/commands/components/test.rs
@@ -4,10 +4,13 @@ use edgee_components_runtime::config::{ComponentsConfiguration, DataCollectionCo
 use edgee_components_runtime::context::ComponentsContext;
 use std::collections::HashMap;
 
+use edgee_components_runtime::data_collection;
+use std::str::FromStr;
+
 use edgee_components_runtime::data_collection::exports::edgee::protocols::data_collection::EdgeeRequest;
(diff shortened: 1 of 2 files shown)


crates/cli/src/commands/components/test.rs
```diff
@@ -212,18 +215,35 @@ async fn test_data_collection_component(opts: Options) -> anyhow::Result<()> {
...
+            .map(|(k, v)| (k.to_string(), v.to_str().unwrap().to_string()))
+            .collect();
         println!(
             "\t{}: {}",
             "Headers".green(),
             serde_json::to_string_pretty(&pretty_headers)?
                 .to_colored_json_auto()?
                 .replace("\n", "\n\t")
         );
+        println!(
```
alexcasalboni, line 242: why do we need this? aren't all the headers already logged as `pretty_headers`?

alexcasalboni (approved):
Looks good 👍  only a doubt about that additional println!

crates/cli/src/commands/components/test.rs
CLEMENTINATOR: it shows either true or false !

CLEMENTINATOR: its an bool output parameter from the wasm

PR #238, opened 2025-02-20
feat: no more moving path

SachaMorard:
### Checklist

### Description of Changes

Event paths rotated with each page view. The idea was interesting, but it added complexity to the SEO management of these URLs. 
In the end, we use the current url to collect an event, which makes data collection even more efficient.

For this, the sdk is bumped to version 5, and we no longer need the edge-path crate.

### Related Issues

No issue related

### DIFF
--- a/crates/server/src/proxy/controller.rs
+++ b/crates/server/src/proxy/controller.rs
@@ -18,10 +18,9 @@ type Response = http::Response<BoxBody<Bytes, Infallible>>;
 
 pub async fn edgee_client_event(ctx: IncomingContext) -> anyhow::Result<Response> {
     let res = http::Response::builder()
-        .status(StatusCode::NO_CONTENT)
+        .status(StatusCode::OK)
         .header(header::CONTENT_TYPE, "application/json")
         .header(header::CACHE_CONTROL, "private, no-store")
-        .header("X-Robots-Tag", "noindex, nofollow")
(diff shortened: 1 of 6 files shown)


crates/server/src/proxy/controller.rs
```diff
@@ -163,7 +159,7 @@ pub fn options(allow_methods: &str) -> anyhow::Result<Response> {
         .header(header::ACCESS_CONTROL_ALLOW_METHODS, allow_methods)
         .header(
             header::ACCESS_CONTROL_ALLOW_HEADERS,
-            "Content-Type, Edgee-Debug",
+            "Content-Type, Edgee-Debug, Authorization",
```
alexcasalboni, line 162: looking at the code, it looks like this is not applied to all resources (only some specific Edgee-owned paths), so we might need to handle OPTIONS responses more broadly to enable the `Authorization` header everywhere. And I wonder if this is OK/acceptable in all situations for customers 🤔

SachaMorard: We discussed this together live, but I'll rephrase it here.
The OPTIONS method is only called to validate the third party data collection endpoint. Its response concerns only the latter, not the rest of the URLs. 
Consequently, accepting the use of the Authorization header only concerns the part that concerns us.

alexcasalboni (approved):

PR #245, opened 2025-02-24
feat(CLI):  enable component icon upload and open login browser 

NicolasGirardot:
### Checklist

### Description of Changes

Describe your changes here

### Related Issues

List related issues here


crates/cli/src/commands/components/push.rs
```diff
@@ -254,3 +297,11 @@ fn convert_manifest_config_fields(manifest: &Manifest) -> Vec<api_types::Configu
         })
         .collect()
 }
+
+fn hash_reader<R: Read>(mut reader: R) -> anyhow::Result<String> {
+    use sha2::{Digest, Sha256};
+
+    let mut hasher = Sha256::new();
+    std::io::copy(&mut reader, &mut hasher)?;
+    Ok(format!("{:x}", hasher.finalize()))
```
KokaKiwi, line 306: Why returning a string here instead of the hash itself?

crates/cli/src/components/manifest.rs
```diff
@@ -107,6 +108,23 @@ impl Manifest {
             );
         }
 
+        if let Some(ref avatar_path) = manifest.component.avatar_path {
+            let valid_extensions = ["png", "jpg", "jpeg"];
```
KokaKiwi, line 112: Why testing for extension here? What if user uses another image format (WebP, JpegXL, etc.)?
What about testing the MIME type using a `file`-like lib instead if we want to check it's actually an image?

alexcasalboni: We do the same by checking the mime type in the API: https://github.com/edgee-cloud/api/blob/76a4f9568579d98c99d584b53abfa17590dad95b/internal/s3/s3.go#L189

(and we only accept jpg and png for now - we can always extend the list of supported image types later on)

crates/cli/src/components/manifest.rs
```diff
@@ -107,6 +108,23 @@ impl Manifest {
...
+            let extension = Path::new(avatar_path)
+                .extension()
+                .and_then(|ext| ext.to_str())
+                .unwrap_or_default()
+                .to_lowercase();
+
+            if !valid_extensions.contains(&extension.as_str()) {
+                anyhow::bail!(
+                    "Invalid avatar path extension '{}', must be one of: {:?}",
```
alexcasalboni, line 121: I would call it "icon" in the error messages ("avatar" is just our internal db field, not sure why)

alexcasalboni (approved):
LGTM 👍 only an error message update 👌

KokaKiwi (approved):

crates/cli/src/commands/components/init.rs
```diff
@@ -76,6 +76,7 @@ pub async fn run(_opts: Options) -> anyhow::Result<()> {
                 command: component_language.default_build_command.to_string(),
                 output_path: std::path::PathBuf::from(""),
             },
+            avatar_path: None,
```
alexcasalboni, line 79: I would actually call it `icon` in the manifest as well

PR #249, opened 2025-02-27
components-runtime: return send events status in future

CLEMENTINATOR:
### Checklist

### Description of Changes

return status of the provider calls

### DIFF
--- a/crates/components-runtime/src/data_collection/debug.rs
+++ b/crates/components-runtime/src/data_collection/debug.rs
@@ -73,6 +73,7 @@ impl DebugPayload {
     }
 }
 
+#[derive(Clone)]
 pub struct DebugParams {
     pub from: String,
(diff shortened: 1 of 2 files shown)


SachaMorard (approved):

SachaMorard (approved):
LGTM! Should we handle the debug infos (sent to the api) within data-collection now?
I think it'd be better to remove this part from `edgee` repo, and I assume you can add a debug boolean field to your new EventResponse struct. WDYT?

CLEMENTINATOR:
> LGTM! Should we handle the debug infos (sent to the api) within data-collection now? I think it'd be better to remove this part from `edgee` repo, and I assume you can add a debug boolean field to your new EventResponse struct. WDYT?

i think a cleanup is indeed needed ! but I'll do it next week !

PR #252, opened 2025-02-27
feat(cli): add Edgee WIT files updating capabilities

KokaKiwi:
### Checklist

### Description of Changes

- Allow user to download Edgee WIT files directly from the `edgee` CLI (`edgee components wit update`)
- Automatically download Edgee WIT files when initializing a new component (`edgee components init`)

### Related Issues

- Resolves EDGEE-492


CLEMENTINATOR:
is it time for the hidden edgee folder @alexcasalboni ?

alexcasalboni:
> is it time for the hidden edgee folder @alexcasalboni ?

yes!

crates/cli/src/components/wit.rs
```diff
@@ -0,0 +1,112 @@
...
+
+    let archive = Archive::new(tarball);
+    let mut entries = archive
+        .entries()?
+        .filter_map(|entry| std::future::ready(entry.ok()))
+        .filter(|entry| std::future::ready(entry.path_bytes().ends_with(b".wit")));
+
+    let wit_path = root_dir.join(".edgee/wit");
+    if wit_path.exists() {
+        tokio::fs::remove_dir_all(&wit_path).await?;
```
alexcasalboni, line 62: this will delete the lock file too, right?

KokaKiwi: Yep but we regenerate it later

crates/cli/src/commands/components/build.rs
```diff
@@ -9,16 +11,19 @@ pub async fn run(_opts: Options) -> anyhow::Result<()> {
...
+    do_build(&manifest, root_dir).await?;
 
     Ok(())
 }
 
-pub async fn do_build(manifest: &Manifest) -> anyhow::Result<()> {
+pub async fn do_build(manifest: &Manifest, root_dir: &Path) -> anyhow::Result<()> {
     use std::process::Command;
 
+    crate::components::wit::should_update(manifest, root_dir).await?;
```
alexcasalboni, line 25: Do we need to do the same for `create`, `test` and `check` commands?

`test` and `check` only operate on the .wasm file so maybe it's not necessary 🤔 

`create` probably needs it since the new `.edgee` folder won't be tracked in git

crates/cli/src/components/wit.rs
```diff
@@ -36,67 +35,62 @@ impl Lock {
...
+    });
 
-    // Create WIT world file
     let category_config = CATEGORY_OPTIONS
         .iter()
         .find(|&config| config.value == manifest.component.category)
         .expect("should have a valid category");
+
+    // Update deps
+    let deps_manifest = format!(
```
alexcasalboni, line 56: why not writing the `deps.toml` to disk as well? that way, the `.edgee/wit` folder is fully compatible with `wit-deps` and eventually experienced user might decide to use `wit-deps` directly in the `.edgee/wit` folder (even if we handle that transparently for them).

alexcasalboni (approved):
LGTM 🚀

PR #260, opened 2025-03-04
CLI UX improvements

alexcasalboni:
### Checklist

### Description of Changes

- CLI copyedit pass & shortering a few prompts
- add highlight/color to output where it makes sense
- standardize on tracing::info instead of println (unless it's a special case)
- add help for Editor use and customization (via env variable)

### DIFF
--- a/crates/cli/src/commands/components/init.rs
+++ b/crates/cli/src/commands/components/init.rs
@@ -1,4 +1,5 @@
 use url::Url;
+use colored::Colorize;
 
 use crate::components::{
     boilerplate::{CATEGORY_OPTIONS, LANGUAGE_OPTIONS, SUBCATEGORY_OPTIONS},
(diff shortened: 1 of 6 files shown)


CLEMENTINATOR (approved):
colors ! lgtm

crates/cli/src/commands/components/init.rs
```diff
@@ -23,29 +24,30 @@ pub async fn run(_opts: Options) -> anyhow::Result<()> {
...
     let component_language = Select::new(
-        "Select the language of the component:",
+        "Select a programming language:",
         LANGUAGE_OPTIONS.to_vec(),
     )
     .prompt()?;
     let component_category = if CATEGORY_OPTIONS.len() == 1 {
-        CATEGORY_OPTIONS[0].clone() // Accès direct car on sait qu'il y a un seul élément
+        CATEGORY_OPTIONS[0].clone() // there is only 1 element
     } else {
```
KokaKiwi, line 33: Can be simplified into this:
```rust
    let component_category = if let [category] = &CATEGORY_OPTIONS[..] {
        category.clone()
    } else {
```

alexcasalboni: Thanks 🙏  I've only touched the comment because it was in French 😆 

if that's ok for you, I will leave this for a future refactoring (since it's already working and today I don't have time to test the init command from scratch)

PR #271, opened 2025-03-11
auth: bypass credentials.toml if login env vars are provided

CLEMENTINATOR:
### Checklist

### Description of Changes

auth: bypass credentials.toml if login env vars are provided

### DIFF
--- a/crates/api-client/src/auth.rs
+++ b/crates/api-client/src/auth.rs
@@ -39,6 +39,18 @@ impl Config {
     }
 
     pub fn load() -> Result<Self> {
+        // if EDGEE_API_URL and EDGEE_API_TOKEN are set, use them
+        // skip using the credentials file
+        if let Ok(api_token) = std::env::var("EDGEE_API_TOKEN") {
+            return Ok(Self {
+                api_token: Some(api_token),
+                url: Some(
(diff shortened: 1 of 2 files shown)


crates/api-client/src/auth.rs
```diff
@@ -39,6 +39,22 @@ impl Config {
     }
 
     pub fn load() -> Result<Self> {
+        // if EDGEE_API_URL and EDGEE_API_TOKEN are set, use them
+        // skip using the credentials file
+        match (
+            std::env::var("EDGEE_API_URL"),
```
alexcasalboni, line 45: I'd expect this to have a good default even when not explicitly provided via env variable 🤔 

In a CI/CD workflow, I'd expect to set up only `EDGEE_API_TOKEN`.

Imho, `EDGEE_API_URL` is something only the Edgee team will ever use (to test in staging) and I wouldn't force anyone to use that env variable.

CLEMENTINATOR: :+1:

CLEMENTINATOR: done

alexcasalboni (approved):
Looks good to me 👍  let's do some testing too 🚀

CLEMENTINATOR:
@KokaKiwi if you want to take a look and test

PR #273, opened 2025-03-11
feat(cli/components): multiple QoL improvements

KokaKiwi:
### Checklist

### Description of Changes

- Use manifest-defined order for the components settings
- Update component metadata on push even when there is no new version to create (for example: updating the description, icon or visibility)
- Raise `wit-deps` default logging level to `warning`
- Add a non-interactive mode to `components push` command

Also add refactoring of the push code to split it into smaller, more maintainable functions

### Related Issues

Resolves EDGEE-532 and EDGEE-533


crates/cli/src/commands/components/push.rs
```diff
@@ -205,25 +152,35 @@ pub async fn run(opts: Options) -> anyhow::Result<()> {
...
+            push_version()
+                .client(&client)
+                .manifest(&manifest)
+                .organization(&organization)
+                .component_slug(&component_slug)
+                .maybe_changelog(changelog),
+        )
+    } else {
+        tracing::info!(
+            "{} already exists in the registry, edgee will only update the component.",
```
alexcasalboni, line 176: "{} already exists in the registry. Only updating component metadata."

crates/cli/src/logger.rs
```diff
@@ -67,13 +67,15 @@ pub fn init_cli() {
     use tracing_subscriber::prelude::*;
     use tracing_subscriber::{fmt, EnvFilter};
 
+    const DEFAULT_DIRECTIVE: &str = "info,wit_deps=warning";
```
alexcasalboni, line 70: @KokaKiwi  I've tested this and I see see
```
INFO fetch `https://github.com/edgee-cloud/edgee-wit/archive/refs/tags/v0.5.0.tar.gz` into `.edgee/wit/deps/edgee`
 ```

when running `edgee component build`.

I did have `EDGEE_LOG` set as `info`, but now I have unset it and the info log still shows up 🤔 

Can you please double-check this?

File: crates/cli/src/components/manifest.rs [hunk 3]
```diff
@@ -1,9 +1,7 @@
-use std::{
-    collections::HashMap,
-    path::{Path, PathBuf},
-};
+use std::path::{Path, PathBuf};
 
 use anyhow::{Context, Result};
+use indexmap::IndexMap as Map;
```
CLEMENTINATOR, line 4: not sure what i think about hiding the type

alexcasalboni: it could get confused with `std::iter::Map`? I'm inclined to just keep it as `IndexMap` to avoid confusion 👍

alexcasalboni (approved):
LGTM 👍

PR #280, opened 2025-03-12
feat(cli): add telemetry

KokaKiwi:
### Checklist

### Description of Changes

- Add telemetry to CLI

### Related Issues

Resolves EDGEE-389


crates/cli/src/commands/components/mod.rs
```diff
@@ -28,5 +28,44 @@ pub type Options = Command;
...
+async fn send_telemetry_event(command: &Command, res: &anyhow::Result<()>) -> anyhow::Result<()> {
+    use std::collections::HashMap;
+
+    use crate::telemetry::Event;
+
+    let command_name = match command {
+        Command::Build(_) => "build",
+        Command::Check(_) => "check",
+        Command::Init(_) => "init",
+        Command::List(_) => "list",
```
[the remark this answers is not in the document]
alexcasalboni: also, it would be interesting to see if they actually used `component` vs `components` 😄

crates/cli/src/commands/components/push.rs
```diff
@@ -9,24 +9,23 @@ use edgee_api_client::{types as api_types, Client};
...
     /// Defaults to the user "self" org
-    pub organization: Option<String>,
+    organization: Option<String>,
 
     #[arg(short, long, id = "PROFILE", env = "EDGEE_API_PROFILE")]
     profile: Option<String>,
 
     #[arg(long, conflicts_with = "private")]
-    pub public: bool,
+    public: bool,
```
alexcasalboni, line 22: @Jer1605 @CLEMENTINATOR I just noticed that all of these push options don't have a documentation line (with `/// XYZ`). Let's try to fix that tomorrow 🚀

crates/cli/src/telemetry.rs
```diff
@@ -0,0 +1,66 @@
...
+
+        let client = Client::builder()
+            .user_agent(format!("edgee/{}", env!("CARGO_PKG_VERSION")))
+            .build()?;
+
+        let payload = serde_json::json!({
+            "data_collection": {
+                "events": [
+                    {
+                        "type": "user",
```
alexcasalboni, line 30: why user?

crates/cli/src/commands/components/mod.rs
```diff
@@ -28,5 +28,44 @@ pub type Options = Command;
 
 pub async fn run(command: Command) -> anyhow::Result<()> {
     crate::logger::init_cli();
-    command.run().await
+
+    let cmd = command.clone();
+
+    let res = command.run().await;
+    let _ = send_telemetry_event(&cmd, &res).await;
```
alexcasalboni, line 35: any chance this could run asynchronously and in a non-blocking way? (to avoid adding latency to the main command execution)

KokaKiwi: The telemetry event is sent after the command has been run so we can also include the result information (Ok/Err), so no choice but to wait for both of them
Unless we want to keep the result information for a more useful error-tracking feature (with Sentry for example), which in that case mean we can run both in parallel

alexcasalboni: Of course, running telemetry after the command results makes sense 👍  I was wondering if we could avoid a blocking await for the telemetry call itself, so we avoid adding additional waiting time to the command execution. But not a big deal, if it complicates everything.

crates/cli/src/commands/components/push.rs
alexcasalboni: FYI I'm taking care of this in another PR

alexcasalboni: Here: https://github.com/edgee-cloud/edgee/pull/284

crates/cli/src/telemetry.rs
```diff
@@ -0,0 +1,73 @@
...
+                .properties(self.properties),
+        );
+        let events = vec![dc::types::EdgeeEventDataCollectionEventsItem::track(track)?];
+
+        let page = dc::types::EdgeeEventPageData::builder()
+            .title(self.title)
+            .url("cli://edgee-cli")
+            .path("/");
+        let context =
+            dc::types::EdgeeEventDataCollectionContext::builder().page(Some(page.try_into()?));
```
alexcasalboni, line 43: why a page event though? aren't we better off sending custom `track` events? 🤔

KokaKiwi: I changed the event to a `track` type but the context has many fields, among them is the `page` field which is apparently required

crates/cli/src/telemetry.rs
```diff
@@ -23,9 +64,25 @@ impl Event {
     pub async fn send(self) -> Result<()> {
         use edgee_api_client::data_collection as dc;
 
+        if std::env::var("EDGEE_TELEMETRY_OPTOUT")
+            .is_ok_and(|value| value != "1" && value != "true")
```
alexcasalboni, line 68: should it be `value == "1" || value == "true"` ?

I mean, we need to return if the env var has a truthy value, right?

crates/cli/src/telemetry.rs
```diff
@@ -23,9 +64,25 @@ impl Event {
...
+
         let client = dc::new()
             .baseurl(TELEMETRY_BASE_URL)
             .debug_mode(true)
+            .with_default_headers(|headers| {
+                use reqwest::header;
+
+                headers.insert(
+                    header::USER_AGENT,
+                    format!("edgee/{}", env!("CARGO_PKG_VERSION"))
```
alexcasalboni, line 90: imho, this be a default event property too 👌

crates/cli/src/telemetry.rs
KokaKiwi: It's an opt-out env var, so we need to check if the user did *not* opt-out :)

crates/cli/src/telemetry.rs
```diff
@@ -1,12 +1,12 @@
...
 use std::future::Future;
+use std::{collections::HashMap, time::Duration};
 
 use anyhow::Result;
 
 use event_builder::{IsComplete, IsUnset, SetProperties, State};
 
 const TELEMETRY_BASE_URL: &str = "https://edgee-cli.edgee.team";
-
+const TELEMETRY_TIMEOUT: Duration = Duration::from_millis(250);
```
alexcasalboni, line 9: I'd consider starting with a higher number here, even up to 1 second (to cover the worst case of slow connections).

crates/cli/src/telemetry.rs
alexcasalboni: don't we need to enter the if only when the env variable is true or 1? in that case, we do nothing

crates/cli/src/telemetry.rs
```diff
@@ -0,0 +1,160 @@
...
+
+Telemetry
+---------
+The Edgee CLI collect usage data in order to help us improve your experience.
+You can opt-out of telemetry by setting the EDGEE_TELEMETRY_OPTOUT environment
+variable to '1' or 'true' using your favorite shell.
+"#;
+
+pub fn setup() -> Result<()> {
+    let state_dir = dirs::state_dir()
```
alexcasalboni, line 20: Looking at the docs, it seems this might not work on MacOS and Windows:
Screenshot 2025-03-19 at 16 36 46

Can't we use the same folder used for the login info?

KokaKiwi: I'm gonna let it fallback to that same folder I think yeah

crates/cli/src/telemetry.rs
```diff
@@ -0,0 +1,160 @@
...
+        .ok_or_else(|| anyhow::anyhow!("state dir is not existent"))?
+        .join("edgee");
+    std::fs::create_dir_all(&state_dir)?;
+
+    let first_run = state_dir.join("first_run.marker");
+    if !first_run.exists() {
+        eprintln!("{TELEMETRY_WARNING}");
```
alexcasalboni, line 27: is `eprintln` doing something special here? does it format the message differently somehow?

crates/api-client/Cargo.toml
```diff
@@ -14,6 +14,7 @@ edition.workspace = true
 anyhow.workspace = true
 bon.workspace = true
 chrono = { workspace = true, features = ["serde"] }
+cookie = { workspace = true, features = ["percent-encode"] }
```
alexcasalboni, line 17: do we still need this?

crates/cli/src/telemetry.rs
```diff
@@ -0,0 +1,161 @@
...
+                .context(Some(context.try_into()?)),
+        );
+
+        let _res = client
+            .collect_event()
+            .body(payload)
+            .send()
+            .await
+            .inspect_err(|err| {
+                dbg!(err);
```
alexcasalboni, line 135: this is ok to merge? is it the same as using `tracing::debug`?

alexcasalboni (approved):
LGTM 👍  Only a couple of minor comments

crates/cli/src/telemetry.rs
```diff
@@ -0,0 +1,161 @@
+use std::future::Future;
+use std::{collections::HashMap, time::Duration};
+
+use anyhow::Result;
+
+use event_builder::{IsComplete, IsUnset, SetProperties, State};
+
+const TELEMETRY_BASE_URL: &str = "https://edgee-cli.edgee.team";
```
CLEMENTINATOR, line 8: is .team the ocrrect url for prod ?

crates/cli/src/telemetry.rs
SachaMorard: @KokaKiwi when will this message appear?

crates/cli/src/telemetry.rs
SachaMorard: And what about the user context @KokaKiwi ?
It seems missing here.

SachaMorard (changes requested):

crates/cli/src/telemetry.rs
KokaKiwi: It will be printed on the first run of the Edgee CLI (for any command, actually, maybe I should put that on components command when thinking about it..)

crates/cli/src/commands/components/mod.rs
```diff
@@ -29,5 +29,7 @@ pub type Options = Command;
 pub async fn run(command: Command) -> anyhow::Result<()> {
     crate::logger::init_cli();
 
+    let _ = crate::telemetry::setup().inspect_err(|err| tracing::debug!("Telemetry error: {err}"));
```
alexcasalboni, line 32: @KokaKiwi  this means we won't be tracking usage for the other commands? I mean `edgee serve`, `edgee update`, etc.

alexcasalboni (approved):
LGTM 👍  before merging, could you please share access to the Amplitude project to double-check how the raw data is showing up?

crates/cli/src/telemetry.rs
```diff
@@ -52,8 +52,36 @@ impl Data {
...
+                self.save()?;
+            }
+        }
+        Ok(())
+    }
+
+    fn save(&self) -> Result<()> {
+        let data_file = STATE_DIR
+            .as_ref()
+            .ok_or_else(|| anyhow::anyhow!("no state dir"))?
```
alexcasalboni, line 79: @KokaKiwi I remember you had implemented a config-dir fallback at some point?

KokaKiwi: Uh yeah indeed, i don't know why it does not appears here i may have forgot to update some parts

KokaKiwi: Ah yeah no it's okay, the fallback is present in `STATE_DIR` actually, it's just that the type is still an `Option`, which means if the value is `None` something really wrong is happening

crates/cli/src/telemetry.rs
```diff
@@ -0,0 +1,272 @@
...
+use anyhow::Result;
+use serde::{Deserialize, Serialize};
+use tokio::sync::Mutex;
+
+use edgee_api_client::{data_collection as dc, types::UserWithRoles};
+use event_builder::{IsComplete, IsUnset, SetProperties, State};
+
+static STATE_DIR: LazyLock<Option<PathBuf>> = LazyLock::new(|| {
+    dirs::state_dir()
+        .or_else(dirs::config_dir)
```
alexcasalboni, line 15: @KokaKiwi never mind, it's here :)

alexcasalboni (approved):

SachaMorard (approved):
LGTM

PR #288, opened 2025-03-22
multi wit version loading

CLEMENTINATOR:
### Checklist

### Description of Changes

enable edgee to load more than one version of world

this works, but code definitively needs a cleanup


crates/cli/src/commands/components/test.rs
```diff
@@ -118,7 +118,7 @@ async fn test_data_collection_component(opts: Options) -> anyhow::Result<()> {
     let instance = context
         .get_data_collection_instance(&component_path, &mut store)
         .await?;
-    let component = instance.edgee_components_data_collection();
+    let component = instance.edgee_components1_0_0_data_collection();
```
alexcasalboni, line 121: shouldn't this be based on the manifest wit-version?

crates/cli/src/commands/components/push.rs
```diff
@@ -83,6 +83,7 @@ pub async fn run(opts: Options) -> Result<()> {
     match super::check::check_component(
         super::check::ComponentType::DataCollection,
         output_path.to_str().unwrap(),
+        "1.0.0",
```
alexcasalboni, line 86: shouldn't this be based on the manifest wit-version?

crates/components-runtime/src/context.rs
```diff
@@ -62,13 +88,25 @@ impl ComponentsContext {
         let data_collection_components = config
             .data_collection
             .iter()
+            .filter(|entry| entry.version == DataCollectionProtocolVersion::V1_0_0)
```
alexcasalboni, line 91: we could just do a match inside the `.map(|entry| ...)`? so all the cases are in the same place (and we can try to automate that)

crates/components-runtime/src/data_collection/mod.rs
```diff
@@ -1,14 +1,26 @@
 mod context;
-mod convert;
 mod debug;
 pub mod logger;
 pub mod payload;
+pub mod version;
+
+pub mod data_collection_1_0_0 {
+    wasmtime::component::bindgen!({
+        world: "data-collection-one",
```
alexcasalboni, line 9: Did you have to use letters instead of numbers to generate code that makes sense? 🤔 

If possible, I'd try to stick to the full semver, like "data-collection-1-0-0" or "data-collection-one-zero-zero"

crates/components-runtime/wit/deps.toml
```diff
@@ -1 +1,3 @@
 edgee = "https://github.com/edgee-cloud/edgee-wit/archive/refs/tags/v1.0.0.tar.gz"
+edgee2 = "https://github.com/edgee-cloud/edgee-wit/archive/refs/tags/v0.5.0.tar.gz"
```
alexcasalboni, line 2: we could use the version here too?

crates/cli/src/commands/components/test.rs
CLEMENTINATOR: yes i just didnt finish the test part

crates/cli/src/commands/components/push.rs
CLEMENTINATOR: ditto, ubnfinished work

crates/components-runtime/src/data_collection/mod.rs
CLEMENTINATOR: i cannot use numbers

crates/components-runtime/src/data_collection/mod.rs
```diff
@@ -179,80 +189,181 @@ pub async fn send_events(
...
-                    component
-                        .call_page(&mut store, &component_event, &component_settings)
-                        .await
-                }
-                Component::EventType::Track => {
-                    component
-                        .call_track(&mut store, &component_event, &component_settings)
+
+            let (headers, method, url, body) = match cfg.version {
+                version::DataCollectionProtocolVersion::V0_5_0 => {
```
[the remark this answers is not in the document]
CLEMENTINATOR: i dont think we can generalise that but we can move them to different files

crates/components-runtime/wit/deps.toml
CLEMENTINATOR: yep, jsut lazyness here

crates/components-runtime/src/data_collection/mod.rs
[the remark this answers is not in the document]
alexcasalboni: Yes, I think it would be handy to keep all the version-related stuff in a single file, so it becomes easier to handle and add new versions in the future.

crates/components-runtime/src/context.rs
CLEMENTINATOR: sadly as the return type is not the same this is not doable

SachaMorard (commented):
Ok, I read it. Impressive!
We'll definitely try to avoid changing our wit too often ;)
but this approach is ok with me

crates/components-runtime/src/context.rs
```diff
@@ -34,7 +62,7 @@ pub fn pre_instanciate_data_collection_component_internal(
 
     let component = Component::from_file(engine, &component_config.file)?;
     let instance_pre = linker.instantiate_pre(&component)?;
-    let instance_pre = DataCollectionPre::new(instance_pre)?;
+    let instance_pre = DataCollectionZeroFiveZeroPre::new(instance_pre)?;
```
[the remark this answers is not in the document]
CLEMENTINATOR: yep, doesnt work :(

crates/cli/src/commands/components/check.rs
```diff
@@ -88,12 +111,20 @@ pub async fn run(_opts: Options) -> anyhow::Result<()> {
                 .output_path
                 .to_str()
                 .context("Output path should be a valid UTF-8 string")?;
-            component_path.to_string()
+            (
+                component_path.to_string(),
+                ComponentType::DataCollection,
```
alexcasalboni, line 116: aren't you going to need a match also for the manifest component type? maybe we should extract the string->type conversion in its own function and just use it everywhere, something like:

```Rust
match component_type.as_str() {
    "data-collection" => ComponentType::DataCollection,
    "consent-mapping" => ComponentType::ConsentMapping,
    _ => anyhow::bail!(
        "Invalid component type: {}, expected 'data-collection' or 'consent-mapping'",
        component_type
    ),
}
```

CLEMENTINATOR: yep

crates/cli/src/commands/components/test.rs
```diff
@@ -116,7 +127,7 @@ async fn test_data_collection_component(opts: Options) -> anyhow::Result<()> {
     let mut store = context.empty_store_with_stdout();
 
     let instance = context
-        .get_data_collection_instance(&component_path, &mut store)
+        .get_data_collection_1_0_0_instance(&component_path, &mut store)
```
alexcasalboni, line 130: this should depend on the WIT version too, right?

crates/components-runtime/src/context.rs
```diff
@@ -8,15 +8,16 @@ use wasmtime_wasi::{IoView, ResourceTable, WasiCtx, WasiView};
...
+use crate::data_collection::version::DataCollectionWitVersion;
 
 pub struct ComponentsContext {
     pub engine: Engine,
     pub components: Components,
 }
 
 pub struct Components {
-    pub data_collection: HashMap<String, DataCollectionPre<HostState>>,
+    pub data_collection_1_0_0: HashMap<String, DataCollectionPre<HostState>>,
```
alexcasalboni, line 20: how does this evolve once we have 2-3 versions to support?

crates/cli/src/commands/components/test.rs
CLEMENTINATOR: yeees

crates/components-runtime/src/context.rs
CLEMENTINATOR: i dont think we have a choice 😂

crates/components-runtime/src/context.rs
```diff
@@ -59,9 +60,10 @@ impl ComponentsContext {
...
 
         // Data collection components
-        let data_collection_components = config
+        let data_collection_1_0_0_components = config
             .data_collection
             .iter()
+            .filter(|entry| entry.wit_version == DataCollectionWitVersion::V1_0_0)
             .map(|entry| {
                 let instance_pre =
                     pre_instanciate_data_collection_component_internal(&engine, entry)?;
```
alexcasalboni, line 69: it looks like `pre_instanciate_data_collection_component_internal` will need to support multiple versions, as it's currently using `DataCollectionPre::new` (from v1.0.0)

crates/components-runtime/wit/deps.lock
```diff
@@ -2,3 +2,8 @@
 url = "https://github.com/edgee-cloud/edgee-wit/archive/refs/tags/v1.0.0.tar.gz"
 sha256 = "f6e11c5cde39ec04b024b1181f45e896f154f7677ff5a904c99f6af5224dc371"
 sha512 = "21b045d0fb82a793a6176f6d9e98dab868563ab7906e3762601e14f06a715e36c965e83e6f5c76cf016af03df48d219b86165a7c18d14284e967d0b837bf99e4"
+
+[edgee2]
```
alexcasalboni, line 6: can be removed?

crates/components-runtime/src/context.rs
alexcasalboni: wouldn't it make more sense to evolve the `ComponentsContext` struct? we'd have 

```Rust
pub struct ComponentsContext {
    pub engine: Engine,
    pub components_1_0_0: Components_1_0_0,
    pub components_1_0_1: Components_1_0_1,
}
```

and then a Components implementation for each version (that can be defined in the version-specific file), like:

```rust
use crate::data_collection::v1_0_0::data_collection::{DataCollection, DataCollectionPre};
pub struct Components_1_0_0 {
    pub data_collection: HashMap<String, DataCollectionPre<HostState>>,
    pub consent_mapping: HashMap<String, ConsentMappingPre<HostState>>,
}
```

and

```rust
use crate::data_collection::v1_0_1::data_collection::{DataCollection, DataCollectionPre};
pub struct Components_1_0_1 {
    pub data_collection: HashMap<String, DataCollectionPre<HostState>>,
    pub consent_mapping: HashMap<String, ConsentMappingPre<HostState>>,
}
```


and so on...

alexcasalboni: this sound easier to handle whenever there is a new WIT version (since we need to take care of all component types, not just data collection)

CLEMENTINATOR: but that would entail that csp and dc are related which they might not be, but i can find a cleaner way

crates/components-runtime/src/context.rs
CLEMENTINATOR: DataCollectionPre is onlu for 1.0.0, since we released without the version

crates/components-runtime/src/context.rs
alexcasalboni: @CLEMENTINATOR that would only mean that the same struct only contains components of the same WIT version (whether it's DC, CMP, or any new type)

crates/components-runtime/src/context.rs
alexcasalboni: Yes, but each version will have its own DataCollectionPre class, right?

crates/components-runtime/wit/deps.lock
CLEMENTINATOR: yeah

crates/cli/src/commands/components/check.rs
```diff
@@ -52,11 +60,14 @@ pub async fn check_component(
...
+            "1.0.0" => {
+                let _ = context
+                    .get_data_collection_1_0_0_instance(component_path, &mut store)
+                    .await?;
+            }
+            _ => anyhow::bail!("Invalid WIT version: {}", component_wit_version),
+        },
         ComponentType::ConsentMapping => {
             let _ = context
                 .get_consent_mapping_instance(component_path, &mut store)
```
alexcasalboni, line 73: shouldn't this be version-specific too?

alexcasalboni: I understand we only have 1 version for now, but I'd start adding the match code, so it's ready to include new versions when available

CLEMENTINATOR: :+1:

CLEMENTINATOR: fixed in [d97d6aa](https://github.com/edgee-cloud/edgee/pull/288/commits/d97d6aa16a60fb3b9f756a8a5d9fe72400bd61be)

crates/cli/src/commands/components/push.rs
```diff
@@ -84,8 +84,13 @@ pub async fn run(opts: Options) -> Result<()> {
 
     // check if the output file is a valid Data Collection component
     match super::check::check_component(
-        super::check::ComponentType::DataCollection,
+        match manifest.component.category {
```
alexcasalboni, line 87: should this support other types already?

alexcasalboni: (same for the `test` command that always calls `test_data_collection_component`)

crates/components-runtime/src/data_collection/versions/v1_0_0/mod.rs
```diff
@@ -0,0 +1,84 @@
...
+
+pub mod data_collection {
+    wasmtime::component::bindgen!({
+        world: "data-collection-v100",
+        path: "src/data_collection/wit",
+        async: true,
+    });
+}
+
+pub fn pre_instanciate_data_collection_component_1_0_0_internal(
```
alexcasalboni, line 22: we can probably remove the `internal` here 😆

crates/cli/src/commands/components/push.rs
CLEMENTINATOR: not handled on api side yet, so we dont have the keys in openapi.json

crates/components-runtime/src/data_collection/versions/v1_0_0/mod.rs
CLEMENTINATOR: the function without internal is defined just below (used in data collection :d)

CLEMENTINATOR: oh mnvm it works

alexcasalboni: yeah, but why do we have two? they do the same thing, with the same inputs and outputs 🤔  I'd expect the "internal" version to be private in the mod, but it's public and used elsewhere - so I'd just merge the two and only use the public (non-internal) version everywhere. Or am I missing something?

crates/components-runtime/src/context.rs
```diff
@@ -130,21 +86,7 @@ impl ComponentsContext {
...
-
-        if instance_pre.is_none() {
-            return Err(anyhow::anyhow!("component not found: {}", id));
-        }
-
-        instance_pre.unwrap().instantiate_async(store).await
-    }
-
-    pub async fn get_consent_mapping_instance(
+    pub async fn get_consent_mapping_1_0_0_instance(
```
alexcasalboni, line 89: should this be moved to `src/consent_mapping/versions/v1_0_0/mod.rs`?

crates/components-runtime/src/data_collection/versions/v1_0_0/convert.rs
```diff
@@ -0,0 +1,142 @@
+use crate::data_collection::convert::{convert_products, convert_properties};
+use crate::data_collection::payload;
+use crate::data_collection::versions::v1_0_0::data_collection::exports::edgee::components::data_collection as DataCollection1_0_0;
```
alexcasalboni, line 3: if we define this as just `DC` we could more easily copy-and-paste this file into a new version, without having to change the version everywhere? wdyt?

crates/components-runtime/src/data_collection/versions/v1_0_0/execute.rs
```diff
@@ -0,0 +1,95 @@
+use crate::config::DataCollectionComponents;
+use crate::context::HostState;
+use crate::data_collection::insert_expected_headers;
+use crate::data_collection::payload::EventType;
+use crate::data_collection::versions::v1_0_0::data_collection::exports::edgee::components::data_collection as Component1_0_0;
```
alexcasalboni, line 5: same here 👍

crates/components-runtime/src/data_collection/wit/deps.toml
```diff
@@ -0,0 +1,2 @@
+edgee-v100 = "https://github.com/edgee-cloud/edgee-wit/archive/refs/tags/v1.0.0.tar.gz"
+edgee-v050 = "https://github.com/edgee-cloud/edgee-wit/archive/refs/tags/v0.5.0.tar.gz"
```
alexcasalboni, line 2: do we still need to keep this?

crates/components-runtime/src/data_collection/versions/v1_0_0/mod.rs
```diff
@@ -0,0 +1,82 @@
...
+    let instance_pre = linker.instantiate_pre(&component)?;
+    let instance_pre = DataCollectionV100Pre::new(instance_pre)?;
+
+    tracing::debug!("loaded new data collection component");
+
+    Ok(instance_pre)
+}
+
+impl ComponentsContext {
+    pub fn pre_instanciate_data_collection_1_0_0_component(
```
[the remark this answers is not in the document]
CLEMENTINATOR: used in data-coll api

crates/components-runtime/src/data_collection/wit/deps.toml
CLEMENTINATOR: nope :thinking:

PR #307, opened 2025-04-15
Auth component

NicolasGirardot:
### Checklist

### Description of Changes

Describe your changes here

### Related Issues

List related issues here


CLEMENTINATOR:
@alexcasalboni i feel like the caching responsability should be on the caller (in our case data-collection repo), and we should not have dynamodb dependency here, wdyt ?

crates/components-runtime/wit/deps/edgee/data-collection.wit
```diff
@@ -106,9 +106,20 @@ interface data-collection {
         body: string,
     }
 
+    record auth-request {
+        method: http-method,
+        url: string,
+        headers: dict,
+        forward-client-headers: bool,
```
SachaMorard, line 113: is `forward-client-headers` it useful?

crates/components-runtime/src/data_collection/auth.rs
```diff
@@ -0,0 +1,146 @@
...
+use std::time::Duration;
+use tracing::error;
+use wasmtime::Store;
+
+use crate::config::DataCollectionComponents;
+use crate::data_collection::exports::edgee::components::data_collection::{
+    Dict, Guest, HttpMethod,
+};
+
+pub async fn get_dynamodb_client() -> Client {
```
SachaMorard, line 14: I think we should cache it in memory by default, and add adapters to it in Dynamo (and Redis/Memcached...)
Edgee proxy is made for a self-hosted use, not sure all the users will be DynamoDB customers

SachaMorard (changes requested):

crates/components-runtime/src/data_collection/auth.rs
alexcasalboni: agreed!

crates/components-runtime/src/data_collection/auth.rs
```diff
@@ -0,0 +1,146 @@
...
+                .await
+        }
+    };
+
+    match res {
+        Ok(res) => {
+            let body = res.text().await.unwrap_or_default();
+            if let Ok(json) = serde_json::from_str::<serde_json::Value>(&body) {
+                if let Some(token) = json
+                    .get(&auth_request.token_response_name)
```
alexcasalboni, line 132: You are assuming that there's only 1 response field to fetch and cache.

For AWS, the field is called "Credentials" and it includes a new access key ID, secret access key, and security (or session) token. We could treat it as a serialized JSON string, so the component will have to deserialize `settings.token` into the correct credentials. But I was hoping that would be a bit more transparent from the component point of view (so the proxy would just over-write the correct settings).

Reference: https://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRole.html#API_AssumeRole_ResponseElements

crates/components-runtime/src/data_collection/mod.rs
```diff
@@ -192,13 +193,37 @@ pub async fn send_events(
...
+                        }
+                    }
+                }
+            } else {
+                None
+            };
+
+            if let Some(ref token) = token {
+                println!("Token fetched from auth: {}", token);
+                component_settings.push(("token".to_string(), token.clone()));
```
alexcasalboni, line 224: what if the component already defines a setting named `token`? I'm not sure this is the best approach 🤔

PR #316, opened 2025-04-25
New Auth process 

NicolasGirardot:
### Checklist

### Description of Changes

Describe your changes here

### Related Issues

List related issues here


crates/components-runtime/src/data_collection/mod.rs
```diff
@@ -54,6 +54,13 @@ pub struct EventResponse {
     pub request: Request,
 }
 
+#[derive(Clone)]
+pub struct AuthResponse {
+    pub component_id: String,
+    pub token: String,
```
CLEMENTINATOR, line 60: what if we need multiple tokens from the response ?

CLEMENTINATOR (commented):
we also could setup the auth process in edgee !

crates/components-runtime/src/data_collection/mod.rs
alexcasalboni: imho, it's challenging to handle all possible cases with multiple tokens, so we might need to force the developer to handle (de)serialization in the component implementation (like for AWS `Credentials` that include 3-4 sub-properties).

crates/components-runtime/src/data_collection/mod.rs
```diff
@@ -78,6 +85,96 @@ pub async fn send_json_events(
...
+                    component_id,
+                    token: String::new(),
+                    token_duration: auth_metadata.token_duration,
+                }
+            }
+        };
+
+        let response_text = match res {
+            Ok(r) => r.text().await.unwrap_or_default(),
+            Err(_) => String::new(),
```
alexcasalboni, line 157: should we log this somewhere too? failing silently is (almost) always a bad idea

crates/components-runtime/src/data_collection/mod.rs
```diff
@@ -78,6 +85,96 @@ pub async fn send_json_events(
...
+            }
+            "POST" => {
+                client
+                    .post(url.clone())
+                    .headers(headers)
+                    .body(body.clone())
+                    .send()
+                    .await
+            }
+            _ => {
```
alexcasalboni, line 146: should we fail earlier (and log it somewhere) if the HTTP method is not supported?

crates/components-runtime/src/data_collection/mod.rs
```diff
@@ -78,6 +85,96 @@ pub async fn send_json_events(
...
+            Err(_) => String::new(),
+        };
+
+        let token = serde_json::from_str::<serde_json::Value>(&response_text)
+            .ok()
+            .and_then(|json| {
+                json.get(&auth_metadata.token_property_name)
+                    .and_then(|v| v.as_str().map(|s| s.to_string()))
+            })
+            .unwrap_or_default();
```
alexcasalboni, line 166: should we log something in case the JSON parsing fails? (this will be really challenging to debug if/when it fails, if we don't log anything and just fail silently)

crates/components-runtime/src/data_collection/mod.rs
```diff
@@ -188,6 +285,22 @@ pub async fn send_events(
                         }
                     }
                 }
+                versions::DataCollectionWitVersion::V1_0_1 => {
+                    match crate::data_collection::versions::v1_0_1::execute::get_edgee_request(
```
alexcasalboni, line 289: shouldn't we handle the auth request here too? it doesn't seem like we're calling `is_auth_required` or `get_auth_request` anywhere?

alexcasalboni: I understand we do it in the DC API, but we should also do the same in the Edgee proxy, right?

crates/components-runtime/src/data_collection/mod.rs
```diff
@@ -78,6 +85,96 @@ pub async fn send_json_events(
...
+    )
+}
+
+pub async fn get_auth_request(
+    context: &ComponentsContext,
+    component: &DataCollectionComponents,
+) -> anyhow::Result<JoinHandle<AuthResponse>> {
+    let mut store = context.empty_store();
+    let (headers, method, url, body, auth_metadata) =
+        match crate::data_collection::versions::v1_0_1::execute::get_auth_request(
```
alexcasalboni, line 108: what happens when v1_0_2 is available? should we match the WIT version here?

crates/components-runtime/src/data_collection/versions/v1_0_1/execute.rs
```diff
@@ -0,0 +1,194 @@
...
+        .get_data_collection_1_0_1_instance(&cfg.id, store)
+        .await
+    {
+        Ok(instance) => instance,
+        Err(err) => {
+            error!("Failed to get data collection instance. Error: {}", err);
+            return Err(err);
+        }
+    };
+    let component = instance.edgee_components1_0_1_data_collection();
```
alexcasalboni, line 117: maybe we should handle the is-auth-required logic directly here? so it's already implemented for both OSS proxy and DC API?

alexcasalboni (dismissed):
A few comments and doubts here and there :)

PR #318, opened 2025-05-07
chore: new component subcategory

SachaMorard:
### Checklist

### Description of Changes

Add a enw component subcategory: server-side

### Related Issues

List related issues here

### DIFF
--- a/crates/api-client/openapi.json
+++ b/crates/api-client/openapi.json
@@ -2633,7 +2633,7 @@
         "required": false,
         "schema": {
           "type": "string",
-          "enum": ["analytics", "warehouse", "attribution", "conversion api"],
+          "enum": ["analytics", "warehouse", "attribution", "conversion api", "server side"],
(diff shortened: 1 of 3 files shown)


alexcasalboni:
What's the "server-side" category for? It seems quite generic and ambiguous. Maybe we can find a better term?

SachaMorard:
> What's the "server-side" category for? It seems quite generic and ambiguous. Maybe we can find a better term?

It's a subcategory of "data collection"
It'll host `meta-signals-gateway` components, and `commanders-act`, and maybe `gtm` :)
Do you have another idea?

alexcasalboni:
@SachaMorard still sounds like analytics to me :) I would expect to see those components next to GA, Amplitude, Segment, etc. Not sure we need an ad-hoc category just because the component uses an official server-side SDK/API (rather than an API typically called by the client, as for the other "analytics" components).

alexcasalboni (approved):

PR #321, opened 2025-05-12
feat: add event filtering and data manipulation

NicolasGirardot:
In this PR we have added a logic to be able to add conditional data manipulation and event filtering 

### Checklist

### Description of Changes

Describe your changes here

### Related Issues

List related issues here

### DIFF
--- a/crates/components-runtime/src/data_collection/mod.rs
+++ b/crates/components-runtime/src/data_collection/mod.rs
@@ -114,9 +114,9 @@ pub async fn send_events(
                 trace_component.is_some() && trace_component.as_ref().unwrap() == cfg.id.as_str();
 
             if cfg.event_filtering_rules.iter().any(|rule| {
-                rule.conditions.iter().any(|condition| {
-                    event.evaluate_filter(&condition.field, &condition.operator, &condition.value)
-                })
+                rule.conditions
+                    .iter()
+                    .any(|condition| event.should_filter_out(condition))
             }) {
                 trace_disabled_event(trace, "event_filtered");
                 continue;
--- a/crates/components-runtime/src/data_collection/mod.rs
+++ b/crates/components-runtime/src/data_collection/mod.rs
@@ -114,9 +114,13 @@ pub async fn send_events(
                 trace_component.is_some() && trace_component.as_ref().unwrap() == cfg.id.as_str();
 
             if cfg.event_filtering_rules.iter().any(|rule| {
-                rule.conditions
+                rule.event_types
                     .iter()
-                    .any(|condition| event.should_filter_out(condition))
+                    .any(|event_type| event_type == &event.event_type.to_string())
+                    && rule
+                        .conditions
+                        .iter()
+                        .any(|condition| event.should_filter_out(condition))
             }) {
                 trace_disabled_event(trace, "event_filtered");
                 continue;
--- a/crates/components-runtime/src/data_collection/mod.rs
+++ b/crates/components-runtime/src/data_collection/mod.rs
@@ -113,6 +113,17 @@ pub async fn send_events(
             let trace =
                 trace_component.is_some() && trace_component.as_ref().unwrap() == cfg.id.as_str();
 
+            if cfg.event_filtering_rules.iter().any(|rule| {
+                rule.conditions.iter().any(|condition| {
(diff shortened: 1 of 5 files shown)


crates/components-runtime/src/data_collection/payload.rs
```diff
@@ -106,6 +108,292 @@ impl Event {
...
+                });
+        });
+    }
+}
+
+pub fn evaluate_boolean_filter(field_value: bool, operator: &str, condition_value: bool) -> bool {
+    match operator {
+        "eq" => field_value == condition_value,
+        "neq" => field_value != condition_value,
+        _ => false,
```
alexcasalboni, line 371: if a wrong operator is used, we might want to log it somewhere because it's probably a bug in the UI or somewhere else :) never fail silently, or this will be un-debuggable in the future

crates/components-runtime/src/data_collection/payload.rs
```diff
@@ -106,6 +108,292 @@ impl Event {
...
+
+pub fn evaluate_string_filter(field_value: &str, operator: &str, condition_value: &str) -> bool {
+    match operator {
+        "eq" => field_value == condition_value,
+        "neq" => field_value != condition_value,
+        "in" => condition_value.split(',').any(|v| v.trim() == field_value),
+        "nin" => !condition_value.split(',').any(|v| v.trim() == field_value),
+        "is_null" => field_value.is_empty(),
+        "is_not_null" => !field_value.is_empty(),
+        _ => false,
```
alexcasalboni, line 383: same

crates/components-runtime/src/data_collection/payload.rs
```diff
@@ -106,6 +108,292 @@ impl Event {
...
+
+pub fn evaluate_number_filter(field_value: &f64, operator: &str, condition_value: &f64) -> bool {
+    match operator {
+        "eq" => field_value == condition_value,
+        "neq" => field_value != condition_value,
+        "gt" => field_value > condition_value,
+        "lt" => field_value < condition_value,
+        "gte" => field_value >= condition_value,
+        "lte" => field_value <= condition_value,
+        _ => false,
```
alexcasalboni, line 395: same

crates/components-runtime/src/data_collection/mod.rs
```diff
@@ -113,6 +113,17 @@ pub async fn send_events(
             let trace =
                 trace_component.is_some() && trace_component.as_ref().unwrap() == cfg.id.as_str();
 
+            if cfg.event_filtering_rules.iter().any(|rule| {
+                rule.conditions.iter().any(|condition| {
+                    event.evaluate_filter(&condition.field, &condition.operator, &condition.value)
```
alexcasalboni, line 118: minor: for coherence with `apply_data_manipulation_rules`, you might want to pass the `condition` object here (instead of the field, operator, and value individually.

crates/components-runtime/src/data_collection/payload.rs
```diff
@@ -106,6 +108,292 @@ impl Event {
 
         all
     }
+    pub fn evaluate_filter(&self, query: &str, operator: &str, value: &str) -> bool {
```
alexcasalboni, line 111: it's not very clear whether the event is filtered or not based on the output of this function.

If the output is true, is the event discarded or not? You might want to clarify that in the function name, or at least add a bit of function documentation.

crates/components-runtime/src/data_collection/payload.rs
```diff
@@ -106,6 +108,292 @@ impl Event {
...
+    }
+    pub fn apply_data_manipulation_rules(&mut self, rules: &[ComponentDataManipulationRule]) {
+        rules.iter().for_each(|rule| {
+            rule.event_types
+                .iter()
+                .for_each(|event_type| match event_type.as_str() {
+                    "page" => {
+                        if let Data::Page(ref mut data) = self.data {
+                            rule.manipulations.iter().for_each(|manipulation| {
+                                let value = data.properties.get(&manipulation.from_property);
```
alexcasalboni, line 320: does this mean that a manipulation rule can only change the page properties? can it not change the page title, for example?

alexcasalboni (approved):
Looks good 👍  only a couple of minor comments!

PR #338, opened 2025-06-10
feat: add real client headers

SachaMorard:
### Checklist

### Description of Changes

Add real client headers

### Related Issues

List related issues here


crates/components-runtime/src/data_collection/versions/v1_0_0/execute.rs
```diff
@@ -14,6 +15,7 @@ pub async fn get_edgee_request(
     component_ctx: &ComponentsContext,
     cfg: &DataCollectionComponents,
     store: &mut Store<HostState>,
+    client_headers: &HashMap<String, String>,
```
alexcasalboni, line 18: Don't we also need to change something here? 🤔  https://github.com/edgee-cloud/edgee/blob/main/crates/components-runtime/src/data_collection/mod.rs#L191

PR #341, opened 2025-06-11
feat: new component categories

SachaMorard:
### Checklist

### Description of Changes

- New component categories and sub categories to prepare the next move
- Remove ConsentMapping components


KokaKiwi (changes requested):

crates/cli/src/components/manifest.rs
```diff
@@ -61,8 +67,19 @@ pub enum SubCategory {
...
     ConversionApi,
-    ConsentMapping,
+    WasmFunction,
     ServerSideTagging,
+    Endpoint,
+    KvStore,
+    BotProtection,
+    RateLimiting,
+    AntiFraud,
+    ConsentMapping,
```
alexcasalboni, line 77: are these worth adding here if they won't be developed using Wasm and the Edgee CLI?

PR #345, opened 2025-06-18
cli: main.rs: return error code to shell

CLEMENTINATOR:
### Checklist

### Description of Changes

Describe your changes here

### Related Issues

List related issues here

### DIFF
--- a/crates/cli/src/main.rs
+++ b/crates/cli/src/main.rs
@@ -6,6 +6,8 @@ mod config;
 mod logger;
 mod telemetry;
 
+use std::process::ExitCode;
+
 #[derive(Debug, Parser)]
 #[command(about, author, version)]
 struct Options {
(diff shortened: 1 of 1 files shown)


KokaKiwi:
I would suggest using [`std::process::ExitCode`](https://doc.rust-lang.org/std/process/struct.ExitCode.html) directly :)

crates/cli/src/main.rs
```diff
@@ -13,7 +13,7 @@ struct Options {
     command: commands::Command,
 }
 
-fn main() {
+fn main() -> Result<(), ()> {
```
KokaKiwi, line 16: I would suggest using [`std::process::ExitCode`](https://doc.rust-lang.org/std/process/struct.ExitCode.html) directly :)

KokaKiwi (changes requested):

KokaKiwi (approved):

PR #375, opened 2025-07-23
outgoing send request handler

CLEMENTINATOR:
- **test: rebuild on launch**
- **impl send_request to mark a component outbound request**


crates/components-runtime/src/context.rs
```diff
@@ -158,6 +159,19 @@ impl WasiHttpView for HostState {
...
+    fn send_request(
+        &mut self,
+        mut request: hyper::Request<wasmtime_wasi_http::body::HyperOutgoingBody>,
+        config: wasmtime_wasi_http::types::OutgoingRequestConfig,
+    ) -> wasmtime_wasi_http::HttpResult<wasmtime_wasi_http::types::HostFutureIncomingResponse> {
+        request.headers_mut().insert(
+            "x-edgee-component-request",
+            HeaderValue::from_static("true"),
+        );
+        Ok(wasmtime_wasi_http::types::default_send_request(
```
alexcasalboni, line 171: can't you actually use the official `default_send_request` here?

crates/components-runtime/src/context.rs
```diff
@@ -158,6 +159,19 @@ impl WasiHttpView for HostState {
     fn ctx(&mut self) -> &mut WasiHttpCtx {
         &mut self.http
     }
+    fn send_request(
+        &mut self,
+        mut request: hyper::Request<wasmtime_wasi_http::body::HyperOutgoingBody>,
+        config: wasmtime_wasi_http::types::OutgoingRequestConfig,
+    ) -> wasmtime_wasi_http::HttpResult<wasmtime_wasi_http::types::HostFutureIncomingResponse> {
```
alexcasalboni, line 166: bonus points if we start using this for logging/tracking/monitoring outgoing requests too, so we have some visibility into what HTTP requests are being sent, to which domains, etc.

crates/components-runtime/src/context.rs
CLEMENTINATOR: its the one though :thinking:

alexcasalboni: never mind, I was too tired 🤣

